PR #15, opened 2015-01-08
Added browserify directive to make sure that JSX->JS transform happens w...

shitalm:
Currently we are getting errors about illegal characters when we use reactify transform on the local module which imports react-autocomplete. This is because reactify works only on local modules and not the imported ones unless we explicitly specify reactify transform in package.json. This patch adds reactify directive and also css directive to package.json

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,31 @@
-examples/build
-node_modules
+# Eclipse
+.classpath
+.project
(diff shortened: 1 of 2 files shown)


ahutchings:
The "reactify" module also needs to be added to dependencies in order for this to work.

PR #16, opened 2015-01-23
Include reactify transform for browserify users

EugeneZ:
There's a webpack configuration file for webpack users, but nothing for browserify. I suppose ideally the npm module would have the built lib instead of the raw one so that all build systems are supported, but in the meantime this is a low-effort change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,5 +21,8 @@
     "autocomplete",
     "combobox",
     "a11y"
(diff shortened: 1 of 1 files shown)


EugeneZ:
Oops, should check before I make a PR -- someone already made the same PR two weeks ago.

PR #22, opened 2015-02-17
Support browserify reactify

alvinsj:


alvinsj:
sorry, this is by mistake

PR #28, opened 2015-03-27
Fix broken link in example.

hellojwilde:
So that it doesn't 404. :D


hellojwilde:
Whoops, I'm the 5th PR for this.

PR #52, opened 2015-09-29
[fixed]: check item length before calling scrollinto view

xujihui1985:
the state of highlightIndex will not be set to null when I change
the value of input box, so exception will be thrown when it is going
to find the highlighted dom element

image


sprjr:
@xujihui1985 think you could get this rebased so we can include it for 1.0.0 in #109?

xujihui1985:
ok. glad to do that

CMTegner:
Superseded by #139.

PR #54, opened 2015-10-22
Allows onBlur function to be set as additional component prop

ejbyne:
This change enables an onBlur function to be set as an additional prop of the component, which is then called via the component's handleInputBlur function (passing the value of the input box as a parameter).

I am currently working on an application where I need this functionality, so that the value of the input box can be processed by the parent component on a blur event. This is a useful addition to the existing onChange prop (called when the user types a character) and onSelect prop (called when an autocomplete option is selected). For my application I specifically need to know the value on the onBlur event so that this value can be saved at that point.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -270,19 +270,6 @@ let Autocomplete = React.createClass({
     return React.cloneElement(menu, { ref: 'menu' })
   },
 
-  getActiveItemValue () {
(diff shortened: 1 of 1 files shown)


CMTegner:
Hi! Thanks for taking the time to create this PR. I'm closing this PR since it'll likely be superseded shortly by the work that will be spun off as a part of #138.

PR #55, opened 2015-10-27
add linting using eslint consuming the rackt config

pwmckenna:
while the style will probably change as more repos are updated (hence the overrides for this repo), this PR introduces eslint and uses the rackt eslint config). I didn't want this to change any code, so rules that did not pass were overridden/disabled to just enforce the style that already exists. 

Here is the repo for the base config being enforced: https://github.com/rackt/eslint-config-rackt

The original discussion around a centralized eslint config can be found here: https://github.com/rackt/history/pull/112


package.json
```diff
@@ -13,14 +13,18 @@
...
+    "test": "npm run lint && echo 'lol'",
+    "start": "rackt server",
+    "lint": "eslint ."
   },
   "authors": [
     "Ryan Florence <rpflorence@gmail.com>"
   ],
   "license": "MIT",
   "devDependencies": {
+    "babel-eslint": "^4.1.3",
```
sprjr: I'd prefer these were pinned explicitly - aka `4.1.3`, I've had a few issues even with devDeps where things break and it's a bit nebulous to tell why.

sprjr:
@CMTegner I rebased this branch off master and resolved the conflicts (not sure why GH is showing that we cannot merge but I can take a look later).

I did actually run `npm install` and `npm run lint` though, and I did not get any errors or warnings at all. That seemed like it would not be the case, but I didn't read the rules yet or the rackt plugin that it extends to see if there actually was validation errors and they weren't caught, or if things were working as intended.

CMTegner:
@sprjr I couldn't get it to work with `babel-eslint@4`, but bumping it to v6 worked. I also upgraded `eslint` to v2, since I felt it was a shame to start a version behind the current stable release. Toggling rules in `.eslintrc` now produces lots of errors. I'll create a ticket where we can track progress towards removing our local overrides.

Just merge this if you think it looks fine.

CMTegner:
Oh, and we should probably lint as a part of build, and set up travis to do it for us.

sprjr:
@CMTegner I was just thinking about a Travis config for this last night when I rebased! I'll open an issue for that as well. I put this in the `1.1.0` milestone for now, as I was going to look at #111 first so we can get the next patch release moving soon.

PR #60, opened 2015-11-10
Fire onSelect event even if there is no match items

tlenclos:
…ere is no match, this change fire the onSelect event on enter keypress


iamdevonbutler:
In the meantime, you can use a simple hack to fire the onSelect callback when the enter key is pressed:

```
<Autocomplete
  inputProps={{
    onKeyPress: (event) => event.key === 'Enter' && ::this.handleSelect(event.target.value),
  }}/>

```

CMTegner:
We'll be adding support for `onKeyPress` and event handlers shortly (#138). I don't think triggering `onSelect` with no item is the way we want to solve this problem. It would also be a breaking change, so it would have to wait for a 2.0 release. I'm closing this issue with the intention of finding a solution in #138.

PR #63, opened 2015-12-03
add missing `event` arg to keyDownHandlers

liamgriffiths:
It looks like this `event` arg was accidentally left out - so this puts it in :smile:


liamgriffiths:
closing this one - noticed that there is already an open PR.

@ryanflorence is this project still maintained?

PR #67, opened 2015-12-30
[added] select item on arrow right key down

Egor-Sapronov:
it's usefull to select item not only on enter but on right arrow key down too


CMTegner:
Thanks! Mind rebasing this?

CMTegner:
After having thought about this a second time I now see that this would be problematic if the user has placed the caret somewhere to the left of the last character. Trying to use the right-arrow to get back to the end of the text would cause an item to be selected.

PR #69, opened 2016-01-15
Call setMenuPositions() on scroll & resize

cdevienne:
If the input position changes, the menu does not follow the move.

This patch fixes it by calling setMenuPosition on resize and scroll, with a throttle of 16msec.


CMTegner:
`setMenuPositions` will likely be removed for v2, so we won't be prioritising any fixes for this at the moment. Regardless, thanks for the effort!

PR #71, opened 2016-01-29
[fixed] handleChange should use event.persist()

kentcdodds:
As noted in [this issue on React](https://github.com/facebook/react/issues/5939), `0.14.7` (released yesterday) introduces some memory leak fixes that result in `event` having most of its properties (like `target` for example) set to `null`.

Calling `event.persist()` will prevent React from removing those properties ([docs](https://facebook.github.io/react/docs/events.html#event-pooling)).

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -88,6 +88,7 @@ let Autocomplete = React.createClass({
 
   handleChange (event) {
     this._performAutoCompleteOnKeyUp = true
+    event.persist()
     this.setState({
       value: event.target.value,
     }, () => {


CMTegner:
LGTM. Just out of curiosity: What are you using `event` for when you receive it via `onChange`? I've been debating for a while if components that wrap `<input>`s should return the entire event object or just the relevant subset (`target.value` in this case).

kentcdodds:
I'm away from my keyboard right now, but I think that's all I'm using it for...

kentcdodds:
That would probably be a separate breaking change/optimization to this one IMO

CMTegner:
Oh definitely. Maybe it's something to consider before 1.0.

sprjr:
@kentcdodds @CMTegner overlooked this one, mind getting it rebased and the conflicts resolved? Not sure I can get this pushed out before I'm out of the country, but would like to include it in the next release candidate for `1.0.0`. Thanks!

kentcdodds:
Actually, looking at it, this is no longer relevant because the API to `onChange` has changed. Everything looks good now :-)

CMTegner:
@kentcdodds are you sure? Still looks like this on master: `onChange(e: Event, value: String)`

kentcdodds:
Ah, yes, you're right. I'll just use the GitHub based editor so I'll close this one. Thanks!

kentcdodds:
Actually, the is not needed any longer because you now invoke the `onChange` callback in sync (rather than as a callback to `setState`). So we no longer need to persist the event :)

CMTegner:
Ah, good point! Nothing to see here :)

PR #74, opened 2016-02-17
Issue41

### LINKED ISSUE #41: text input is missing a visible label and an accessible name
WCAG requires labels to always be visible and each input needs an accessible name, those can be the same if using a <label> element

ryanalane:
Resolves #41


CMTegner:
Looks like these changes have already been merged. Can this PR be closed?

PR #75, opened 2016-02-22
Set the position of menu as "absolute"

chenzhihao:
Set the position of menu as "absolute" so the menu can scroll and position correctly in dialog/modal.

I just changed the src file. Can't find build script so I didn't compile it to build/dist folder.


ztratar:
Would love to see this soon! This is my #1 issue with this library right now -- looks very low quality for users if they happen to scroll.

CMTegner:
You can override all of the menu styles, either via `props.menuStyle` or by applying styles directly in the `props.renderMenu` function.

ztratar:
But left and top are calculated dynamically and will override most cases,
no?

CMTegner:
You don't have to apply all the styles, you can choose which to apply in `props.renderMenu`. You don't have to use `props.menuStyle` at all (or the `style` parameter passed to `renderMenu`).

a-r-d:
@CMTegner I would like to see this merged, or at least the other version here:
https://github.com/reactjs/react-autocomplete/pull/135 Let me know if there is anything I can do to help. 

The position fixed + calculating a top position + setting a max height ends up being pretty awkward in practice. We ended up having to do this on our project to deal with the usability issues. Hope this helps someone:

```css
someSelectorForTheMenuComponentGoesHere {
   position: absolute !important;
   max-height: auto !important;
   top: auto !important;
}
```

PR #76, opened 2016-02-22
[added]add limitation of menu result item amount

chenzhihao:
Show all the search results in menu is a bit silly. Can we add this feature to limit the amount of menu item?

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -12,7 +12,8 @@ let Autocomplete = React.createClass({
     shouldItemRender: React.PropTypes.func,
     renderItem: React.PropTypes.func.isRequired,
     menuStyle: React.PropTypes.object,
-    inputProps: React.PropTypes.object
+    inputProps: React.PropTypes.object,
+    maxItem: React.PropTypes.number,
   },
 
   getDefaultProps () {
@@ -20,7 +21,10 @@ let Autocomplete = React.createClass({
       inputProps: {},
       onChange () {},
       onSelect (value, item) {},
-      renderMenu (items, value, style) {
+      renderMenu (items, value, style, component) {
+        if (component.props.items.length > component.props.maxItem) {
+          items.push(<div>Keep typing to refine results</div>)
+        }
         return <div style={{...style, ...this.menuStyle}} children={items}/>
       },
       shouldItemRender () { return true },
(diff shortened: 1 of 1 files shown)


NerdHerd91:
I would think this is something your api call should be limiting for you on the server-side results.

chenzhihao:
I think this is a handy feature to do limitation in frontend. And also you can leave this to backend, just do not set the props "maxItem".

CMTegner:
It's better to handle this in your application code, since you already need to add logic to limit the amount of items you fetch. You already pass `items` in as a prop, so it just a matter of truncating the list beforehand.

mihaibogdan10:
What about the case when we use static data, though?

davis:
is there a way to do this for static data? rendering all results is less than optimal

CMTegner:
@mihaibogdan10 not sure how it would be any different for static data. Could you provide an example where you think this would be a problem?

davis:
let's say i have a static list of, say the 50 states. `items = ['Alabama', 'Arizona', ..., 'Wyoming']`. how can i apply a limit to when i type in the letter 'a' so only 5 results come back? which part of the API can I use?

CMTegner:
@davis presumably you're already implementing [`shouldItemRender()`](https://github.com/reactjs/react-autocomplete/#shoulditemrender-function-optional) to filter the items based on the input value, so you could move that logic to a function that decides what is contained in `items`:

```jsx
const items  = [...]
const maxItems = 5

class MyAutocomplete extends Component {

  state = { value: '' }

  filterAndLimitItems () {
    return items
      // Only display items that begin with the input term
      .filter(item => item.indexOf(this.state.value) === 0)
      // Limit items
      .slice(0, maxItems)
  }

  render () {
    return (
      <Autocomplete
        onChange={(e, value) => this.setState({ value })}
        items={this.filterAndLimitItems()}
        ...
      />
    )
  }
}
```

You could also implement [`renderMenu()`](https://github.com/reactjs/react-autocomplete/#rendermenu-function-optional) and limit the amount of items it renders:

```jsx
function renderMenu (items, value, style) {
  return (
    <div
      style={{...style, ...this.menuStyle}}
      children={items.slice(0, maxItems)}
    />
  )
}
```

We're still planning API changes for v2 and v3, and one potential change could be to add an `index` parameter to `shouldItemRender()` so you could return false if `index > maxItems`. This isn't as performant as manually limiting `items` since you can't bail out early when you've reached the maximum desired items.

PR #79, opened 2016-03-03
[added]  new props: onFocus and onBlur

matttti:
Adding those props can be useful in many way. For example, in on of my projects, i want to trigger a special ajax search once the input got focus for the first time. Sure, there are other ways to archive this, but they all are somewhat inconvenient :-)


bekerov:
+1

CMTegner:
Any chance you can add tests to prevent accidental regression in this functionality? You can rebase against master to get `__tests__/Autocomplete-test.js` and the test harness.

matttti:
Ok, i added some ...

lib/__tests__/Autocomplete-test.js
```diff
@@ -11,6 +12,10 @@ import Autocomplete from '../Autocomplete';
 import { getStates, matchStateToTerm, sortStates, styles } from '../utils'
 
 chai.use(chaiEnzyme());
+chai.use(spies);
+
+var onFocusSpy = chai.spy()
+var onBlurSpy = chai.spy()
```
CMTegner: These spies don't seem to be reset between each test block, so right now they're only suitable for one-time use. Perhaps a better solution would be to pass them in via `extraProps` to `AutocompleteComponentJSX`, paired with `beforeEach` to reset them for each test test within your block.

lib/__tests__/Autocomplete-test.js
```diff
@@ -41,12 +48,14 @@ describe('Autocomplete acceptance tests', () => {
 
     expect(autocompleteWrapper.state('isOpen')).to.be.false;
     expect(autocompleteWrapper.instance().refs.menu).to.not.exist;
+    expect(onFocusSpy).to.not.have.been.called()
```
CMTegner: Once you get the spy harness sorted you shouldn't have to check their initial state every time.

matttti:
ok, changed that and now pass the spies via extraProps including resets in beforeEach().

lib/__tests__/Autocomplete-test.js
```diff
@@ -11,18 +12,22 @@ import Autocomplete from '../Autocomplete';
...
 chai.use(chaiEnzyme());
+chai.use(spies);
+
+var onFocusSpy = chai.spy()
+var onBlurSpy = chai.spy()
 
 function AutocompleteComponentJSX (extraProps) {
   return (
-    <Autocomplete 
-      initialValue=''
```
CMTegner: What happened here?

lib/__tests__/Autocomplete-test.js
```diff
@@ -88,7 +96,7 @@ describe('Autocomplete kewDown->ArrowDown event handlers', () => {
 
   it('should highlight the "n+1" item in the menu when "n" is selected', () => {
     autocompleteWrapper.setState({'isOpen': true});
-
```
CMTegner: There is a lot of noise in your last commit. Please revert and ensure you're only touching code that relates to this PR. Whitespace cleanup can be done in a separate PR.

CMTegner:
>ok, changed that and now pass the spies via extraProps including resets in beforeEach().

Are you sure? I can't see where the spies are being reset.

matttti:
uh no, i accidentally committed the lines i intended to discard. And vice versa discarded my actual changes! Sorry for that, that was my first try to work with GitHub Desktop. The "real" commit will come soon.

matttti:
...wow, that was the last time i used that tool, what a mess! Please take a look at the changeset of the PR, should be acceptable now. And sorry again for the flood of commits.

sprjr:
I think we should add some examples as well, either that *or* find a place to document all these props and what they do. Thoughts?

sprjr:
I think we should add some examples as well, either that *or* find a place to document all these props and what they do. Thoughts?

NerdHerd91:
This would definitely be needed since when clicking between multiple autocompletes on a page, the previous suggestions will display due to the handleInputFocus. We need the ability to clear from the store when onBlur/onFocus occurs so that the next type of autocomplete starts fresh. 
(you could clear them during onSelect but this won't help if the user chooses not to select one of the suggestions)

CMTegner:
>This would definitely be needed since when clicking between multiple autocompletes on a page, the previous suggestions will display due to the handleInputFocus.

Sounds like a bug to me. Perhaps we should fix this first, then see if this PR is still needed?

lioramsalem90:
+1

CMTegner:
@lioramsalem90 please use the "Add your reaction" feature instead of adding "+1" comments, since they add unwanted noise to the discussion :)

Urthen:
If I'm interpreting @NerdHerd91 correctly, I think he simply means if you have multiple Autocompletes on the same page, depending on how you're storing their state, you might need to determine when one blurs to clear the state (if you share it).

I don't think that's actually a bug, you just need to store the states separately, or do what I did and wrap Autocomplete with a separate state-handling component. It's understandable why it isn't included in Autocomplete as various projects may have differing state handling needs. 

That said, I also need onBlur and onFocus props available for proper validation. This PR seems a bit old and conflicts, can it get updated and merged?

erichhsun:
I've updated this pull request here: https://github.com/reactjs/react-autocomplete/pull/123

CMTegner:
Closing since #123 seems to supersede it.

Issue #25, opened 2015-03-08
Trigger event with query when Enter key is pressed


cleercode:
I'd like to be able to retrieve the query even when the Enter key is pressed, even when an autocomplete option isn't selected, so that the user can submit queries that aren't in the autocomplete options.

emilianosanti:
HI @cleercode! I have a PR for this improvement. I think its functionality is adaptable to your issue.  https://github.com/reactjs/react-autocomplete/pull/83

Badestrand:
I really would like that functionality, too. In my scenario the user can select from a number of options but should also be able to add new items. At least in some cases so the discussed "onSearch" handler would be optimal.

Any chance that pull request might get merged?

vincentbello:
I would argue that this is input behavior, not autocomplete behavior. You can easily handle this use case by catching "enter" keypresses in the input's `onKeyUp` handler and performing your logic with the input value.

jaketlarson:
@vincentbello I'm not sure that's true. I tried something like this, the value of `this` (e.g., `this.state`) suddenly seems to be out of sync in the callback function.

PR #65, opened 2015-12-17
Stop keeping value in state

CMTegner:
a.k.a. replace `props.initialValue` with `props.value`. This change makes the
behaviour of Autocomplete closer to that of native input elements, i.e. you are
forced to always supply the current value which the input should display, and
the element will provide the updated value via the `onChange` callback (and
`onSelect`).

Upgrade path:

Before:
```js
class MyWidget extends Component {
  render() {
    return (
      <Autocomplete
        initialValue="hello"
      />
    )
  }
}
```

After:
```js
class MyWidget extends Component {
  constructor() {
    this.state = { value: 'hello' }
  }
  render() {
    return (
      <Autocomplete
        value={this.state.value}
        onChange={event => this.setState({ value: event.target.value })}
        onSelect={value => this.setState({ value })}
      />
    )
  }
}
```

This is obviously a breaking change.

Closes #49

PR #83, opened 2016-03-23
Trigger event on enter #25

emilianosanti:
The Autocomplete component will receive a function as a propType that will be called when the following events occurs.

- The user enter an element that not exists in the Autocomplete source list.
- The user press enter key.

The function receives a parameter which is the value entered on the input. Thus, who instantiate the Autocomplete component could make any action with that value. For this specific issue, we need to make a service call with this value.

By implementing this improvement with this approach we reach the next benefits:

- This feature is totally optional, in the case we wont add this functionality the component will have the default behavior.
- According to the issue description, we need to make a service call, but if the business rules change (for instance, show an error message highlighting that the search doesn't match with the list) we only need to change the implementation of the function we receive by propType, so the component is totally flexible and adaptable.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -10,7 +10,7 @@ let Autocomplete = React.createClass({
     initialValue: React.PropTypes.any,
     onChange: React.PropTypes.func,
     onSelect: React.PropTypes.func,
-    onSearch: React.PropTypes.func,
+    onSubmit: React.PropTypes.func,
     shouldItemRender: React.PropTypes.func,
     renderItem: React.PropTypes.func.isRequired,
     menuStyle: React.PropTypes.object,
@@ -146,8 +146,8 @@ let Autocomplete = React.createClass({
         this.setState({
           isOpen: false
         }, () => {
-          if (this.props.onSearch)
-            this.props.onSearch(this.state.value);
+          if (this.props.onSubmit)
+            this.props.onSubmit(this.state.value, this.setValue);
           else
             this.refs.input.select();
         })
@@ -176,6 +176,12 @@ let Autocomplete = React.createClass({
         isOpen: false
       })
     }
+  },
+
+  setValue (newValue) {
+    this.setState({
+      value: newValue
+    })
   },
 
   getFilteredItems () {
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -10,6 +10,7 @@ let Autocomplete = React.createClass({
     initialValue: React.PropTypes.any,
     onChange: React.PropTypes.func,
     onSelect: React.PropTypes.func,
+    onSearch: React.PropTypes.func,
     shouldItemRender: React.PropTypes.func,
     renderItem: React.PropTypes.func.isRequired,
     menuStyle: React.PropTypes.object,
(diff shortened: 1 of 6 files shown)


lib/Autocomplete.js
```diff
@@ -10,6 +10,7 @@ let Autocomplete = React.createClass({
     initialValue: React.PropTypes.any,
     onChange: React.PropTypes.func,
     onSelect: React.PropTypes.func,
+    onSearch: React.PropTypes.func,
```
emilianosanti: Feel free to suggest another name for this props.

CMTegner:
What would you do as a result of `onSearch()`? Fetch more items? Perhaps this functionality would be better suited in a `onSubmit()` (form) handler.

emilianosanti:
> What would you do as a result of onSearch()? Fetch more items?

You can do whatever you need inside onSearch function. For instance, it's a improvement referenced to the issue https://github.com/reactjs/react-autocomplete/issues/25, in this function you can submit queries that aren't in the autocomplete options.. 
I think onSubmit() name would it be better.. I will update that

loverajoel:
- Keep the focus of the PR (add an event after Enter key press), if you think that examples requires a refactor please send another PR for that and don't add noise to this PR. cc @emilianosanti 
- I think that this event can be useful in order to set the value from outside the module. ex:
```js

onSearch => (currentValue, setValueCallback) {
	setValueCallback(currentValue.split(',').reverse().join());
}

```
-

emilianosanti:
> Keep the focus of the PR
You're right with that, I'll revert the commit :+1: 

Regarding to the event, it's a great idea, I'll update

CMTegner:
Let's take a step back and review this proposal:

1. Add an API that lets you provide a callback which should be invoked when the user hits return in the input when no suggestion is highlighted.
2. Provide said callback with a way of updating `state.value`.

I'd argue that the first point is the responsibility of the surrounding form, not the input element. Remember, there's already a default behaviour attached to hitting return in an input element, which is to submit the form (naming the callback `onSubmit` makes things even more confusing). Forcing your users to learn a new behaviour for one specific input field probably won't lead to great UX.

As for the second point, there's already work in progress (#65) on removing `state.value` and making `Autocomplete` align with how `value` is treated in regular `<input>` elements (i.e. you control it 100% by passing it as a prop).

I still don't understand the need for this API. No one has offered an example where this proposed callback would be any better than simply using `onChange`. Perhaps I'm missing some important piece of insight into this problem area?

emilianosanti:
I agree with several thing you mentioned, this new API has emerged from an issue requested by cleercode (https://github.com/reactjs/react-autocomplete/issues/25). Do you really think we need this functionality?. I have in mind to decline this pull request

CMTegner:
I honestly don't think it's needed, but I'm not the correct person to argue for its existence since I don't know the problem well enough.

garrettmaring:
I don't see how this is easily implemented outside of the library. To restate the goal there are two desired actions:

- "enter" when an item is highlighted -> trigger onSelect function
- "enter" when no item is highlighted -> trigger some other function

@CMTegner you mentioned that it is the job of the surronding form, suggesting that <Autocomplete /> is wrapped in a form? But that along with other options (assigning listener to the search event if making input type search, or listening for enter key) because difficult since the focus remains on the input field while using arrow keys to navigate through the suggested results.

How can we tell whether an item is highlighted? Or how can we watch for an event that will only fire if no event is highlighted given that all "onSubmit" events will fire if enter is press with or without something selected?

PR #88, opened 2016-04-03
[added] Remove lodash dependency and allow consumers to specify their own id

bfink13:
Lodash is massive and quite unnecessary for generating something as simple as a unique ID. The minified build of react-autocomplete is 72kb (!!!!!!). Besides, consumers should be able (and I'd argue responsible) for supplying an ID.

The alternative to an ID was putting the input inside the label, but I figured that wouldn't be desirable given that it could possibly break some layouts.

### DIFF
--- a/examples/async-data/app.js
+++ b/examples/async-data/app.js
@@ -25,6 +25,7 @@ let App = React.createClass({
         </p>
 
         <Autocomplete
+          id="state-autocomplete"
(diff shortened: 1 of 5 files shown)


NerdHerd91:
+1

bfink13:
Dang it, shouldn't have made this off master. I'll open a new one.

PR #50, opened 2015-09-18
[added] wrapperProps and wrapperStyle props

zoeabryant:
Thank you for this contribution, it has been very useful in our project.
I have created this pull request because we needed more control over the rendering of the widget. Our project has a very strict design manual, so it is important that we are able to add classes and overwrite default styling.

PR #89, opened 2016-04-03
[removed] Remove lodash dependency and hardcoded label

bfink13:
Lodash is massive and quite unnecessary for generating something as simple as a unique ID. The minified build of react-autocomplete is 72kb (!!!!!!). Besides, consumers should be able (and I'd argue responsible) for supplying an ID.

The alternative to an ID was putting the input inside the label, but I figured that wouldn't be desirable given that it could possibly break some layouts.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -1,5 +1,4 @@
 const React = require('react')
-const lodash = require('lodash')
 const scrollIntoView = require('dom-scroll-into-view')
 
 let _debugStates = []
@@ -14,7 +13,6 @@ let Autocomplete = React.createClass({
     renderItem: React.PropTypes.func.isRequired,
     menuStyle: React.PropTypes.object,
     inputProps: React.PropTypes.object,
-    labelText: React.PropTypes.string,
     wrapperProps: React.PropTypes.object,
     wrapperStyle: React.PropTypes.object
   },
@@ -27,7 +25,6 @@ let Autocomplete = React.createClass({
         display: 'inline-block'
       },
       inputProps: {},
-      labelText: '',
       onChange () {},
       onSelect (value, item) {},
       renderMenu (items, value, style) {
@@ -55,7 +52,6 @@ let Autocomplete = React.createClass({
   },
 
   componentWillMount () {
-    this.id = lodash.uniqueId('autocomplete-'); 
     this._ignoreBlur = false
     this._performAutoCompleteOnUpdate = false
     this._performAutoCompleteOnKeyUp = false
@@ -313,11 +309,9 @@ let Autocomplete = React.createClass({
         state: this.state
       })
     }
+
     return (
       <div style={{...this.props.wrapperStyle}} {...this.props.wrapperProps}>
-        <label htmlFor={this.id} ref="label">
-          {this.props.labelText}
-        </label>
         <input
           {...this.props.inputProps}
           role="combobox"
@@ -330,7 +324,6 @@ let Autocomplete = React.createClass({
           onKeyUp={(event) => this.handleKeyUp(event)}
           onClick={this.handleInputClick}
           value={this.props.value}
-          id={this.id}
         />
         {this.state.isOpen && this.renderMenu()}
         {this.props.debug && (
--- a/examples/async-data/app.js
+++ b/examples/async-data/app.js
@@ -16,17 +16,15 @@ let App = React.createClass({
     return (
       <div>
         <h1>Async Data</h1>
-
         <p>
           Autocomplete works great with async data by allowing you to pass in
           items. The <code>onChange</code> event provides you the value to make
           a server request with, then change state and pass in new items, it will
           attempt to autocomplete the first one.
         </p>
-
+        <label htmlFor="states-autocomplete">Choose a state from the US</label>
         <Autocomplete
-          labelText="Choose a state from the US"
-          inputProps={{name: "US state"}}
+          inputProps={{name: "US state", id: "states-autocomplete"}}
           ref="autocomplete"
           value={this.state.value}
           items={this.state.unitedStates}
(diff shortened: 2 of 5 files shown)


CMTegner:
I am 100% for removing the lodash dependency. I <3 lodash, but including the entire library for one function is super overkill. We could include only the parts we need (i.e. `import uniqueId from 'lodash/uniqueId'`), but that's beside the main point: why are we rendering a label at all?

It seems like the `<label>` markup was added without any [discussion around the subject](https://github.com/reactjs/react-autocomplete/pull/77), probably because it was baked into a PR which was really useful (adding tests).

The problem with adding more markup is that the consumers will inevitably want to change the way it's rendered, add props, do various hacks, which means we need to add more code, bloat the API, etc. (we already have this problem with the `<div>` that wraps the autocomplete input and menu, see #91).

I say we remove the `<label>` markup entirely, and let the consumer add an id via `props.inputProps.id` (to link it with their custom `<label>`).

@bfink13 will you modify this PR, or do you want me to create one that supersedes yours?

bfink13:
That's an excellent idea! I'll update the PR.

bfink13:
@CMTegner saw you linked an update where you remove the label and lodash - do you want me to close this one?

CMTegner:
Oh, I didn't realise it got cross referenced here. I just created that branch in case you didn't find the time to do it (I'm hoping we can get it merged before 1.0). If you want me to create a PR from it that's fine by me, I just didn't want to rob you of the chance to fix it :)

bfink13:
Thanks for the opportunity :)

I just updated the PR. There were some conflicts with master on the wrapper <div> but nothing big.

lib/Autocomplete.js
```diff
@@ -313,11 +311,9 @@ let Autocomplete = React.createClass({
         state: this.state
       })
     }
-    return (
+
+    return (
```
CMTegner: Trailing whitespace?

lib/Autocomplete.js
```diff
@@ -313,11 +311,9 @@ let Autocomplete = React.createClass({
         state: this.state
       })
     }
-    return (
+
+    return ( 
       <div style={{...this.props.wrapperStyle}} {...this.props.wrapperProps}>
-        <label htmlFor={this.id} ref="label">
-          {this.props.labelText}
```
CMTegner: Remember to remove `labelText` from `propTypes` and `defaultProps`.

CMTegner:
I think you can just squash all the commits in this PR. It makes it easier to see what actually happened without having having to understand the additional commits.

bfink13:
Alright I think this is all sorted now. This is my first PR through Github, let me know if I need to do anything different with the squashing.

lib/Autocomplete.js
```diff
@@ -313,11 +311,9 @@ let Autocomplete = React.createClass({
         state: this.state
       })
     }
+
     return (
       <div style={{...this.props.wrapperStyle}} {...this.props.wrapperProps}>
-        <label htmlFor={this.id} ref="label">
-          {this.props.labelText}
```
CMTegner: `labelText` still needs removing from `propTypes` and `defaultProps`.

CMTegner:
The easiest way to squash is to `git reset --soft <ref>`, `ref` being the point where you diverged from master. When you've been working on a branch for a while it tends to be a good idea to rebase, to make sure your branch doesn't contain any merge conflicts against master. What I usually do:

* `git rebase upstream master` (I usually set `upstream` to point to the repo I forked from)
* Fix any merge conflicts
* `git reset --soft <ref>`
* `git commit -m 'A sensible commit message'`

I prefer rebasing over merging master into my working branches, since it gives the least amount of commit noise.

bfink13:
Alright I think this should actually be good to go now. Thanks for the tips!

bfink13:
@CMTegner is there a schedule for the next release? When are all these open PRs getting pulled in?

CMTegner:
I'm sure the people with commit access will get around to merging this as soon as they have time :)

bfink13:
LOL geez....with the amount you were participating I figured you were the guy :D my bad.

akhaku:
Just want to add my +1 for this - ran into the issue fixed by #95 earlier, but followed the reference to this PR which is even better. Hope this gets merged into something I can fetch from npm soon!

adamblank:
In addition to being unnecessary, including Lodash like this _breaks _ browserified JS environments that use [underscore.js](http://underscorejs.org/). This is in part due to [how Lodash works with Browserify](https://github.com/lodash/lodash/issues/1852). If Lodash isn't going to be removed from this project, it at least use [noConflict](https://lodash.com/docs#noConflict)

Davorak:
The unique id generation is also causing preventing me for easily doing some equality tests between pages generated in slightly different environments.

So one more vote and one more reason to merge this in.

sprjr:
Good stuff everyone. Sorry for the delays here, I'll get this merged in and publish `1.0.0-rc2`, @CMTegner would like to get an idea of what _else_ we should merge in and finalize for `1.0.0` too.

CMTegner:
@sprjr good to have you back! Here's my take on v1:

Blockers
* #94

Nice-to-have (bugfixes, changes, improvements)
* #52
* #68
* #84
* #92
* #93
* #103
* #104
* #106

There's a couple of topics we need to address either before v1 or soon thereafter:
* Menu styling/positioning: The current implementation and default setup is not great. I'm tempted to just ditch it all and force the user to decide how to render the menu (we'll provide offset and other meta data). The alternative is to go for a relatively positioned menu, which we could easily do since we wrap both the input and the menu with a div. I could mock up both suggestions unless you have a better suggestion. Here are some issues for reference: #69, #70, #75, #100, #102
* `highlightedIndex`: #52 and #106 try to fix some of the issues, but the fact is that `highlightedIndex` isn't updated when `props.items` changes. This needs to be fixed properly.

To be closed (already fixed, no longer applicable, etc)
* #13
* #15
* #30
* #45
* #50
* #53
* #64
* #74
* #95
* #97
* #99
* #105
* + more, these were just the obvious ones

PR #92, opened 2016-04-06
[changed] Prevent `Return` in menu from submitting surrounding form

CMTegner:
The default behaviour of hitting `Return` in a `<input type=text>` is to submit
the immediate surrounding form. I would hazard a guess that this isn't the
expected result when using `Return` to select an item in the autocomplete
menu. Since we don't expose the event object to the consumer via `onSelect` we
need to make sure we're making the most sensible choice by default.

Hitting `Return` a second time (after the menu has closed itself) will
result in the default, expected behaviour.


bfink13:
Looks good!

PR #93, opened 2016-04-09
[changed] Disable native autocomplete

CMTegner:
You can do this via `inputProps`, but since the whole point of this component
is to provide a custom autocomplete solution it makes sense to disable the
native counterpart. If this proves to be problematic we can change the ordering
so that the consumer can override it via `inputProps: { autoComplete: 'on' }`.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -322,6 +322,7 @@ let Autocomplete = React.createClass({
           {...this.props.inputProps}
           role="combobox"
           aria-autocomplete="both"
+          autoComplete="off"
           ref="input"
           onFocus={this.handleInputFocus}
           onBlur={this.handleInputBlur}


lib/Autocomplete.js
```diff
@@ -322,6 +322,7 @@ let Autocomplete = React.createClass({
           {...this.props.inputProps}
```
bfink13: Later props take precedence: https://facebook.github.io/react/docs/jsx-spread.html#spread-attributes

This should come at the end of the props list so that the consumer can override autoComplete (and everything else for that matter).

CMTegner: As I said in the description: 

>If this proves to be problematic we can change the ordering so that the consumer can override it via inputProps: { autoComplete: 'on' }.

Some props we don't want to let the user override (e.g. `onFocus`, `onBlur`), some we *probably* don't want them to override (e.g. `role`, `aria-autocomplete`), while others we *may* want to let them override (e.g. `autoComplete`). I don't know if there has been put any thought into this, so I just followed the existing convention of putting "our" props last.

bfink13: Bah, I totally misread the original description, sorry about that. Sounds good.

CMTegner: Np :)

PR #94, opened 2016-04-09
Dependency cleanup

CMTegner:
1. Demote React to a peer dependency. Having React as a direct dependency can cause multiple versions of React to be bundled in applications that depend on a React version that doesn't match `^0.14.7`.
2. Allow both 0.14 and 15 versions of React
3. Remove duplicate babel-preset-es2015 dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,6 @@
     "dom-scroll-into-view": "1.0.1"
   },
   "peerDependencies": {
-    "react": "^0.14.7"
+    "react": "^0.14.7 || ^15.0.0-0"
   }
 }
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -46,10 +47,10 @@
...
   "dependencies": {
-    "babel-preset-es2015": "^6.5.0",
     "dom-scroll-into-view": "1.0.1",
-    "lodash": "^4.5.0",
-    "react": "^0.14.7",
-    "react-dom": "^0.14.7"
+    "lodash": "^4.5.0"
+  },
+  "peerDependencies": {
+    "react": "^0.14.7 || ^15.0.0-0"
```
bfink13: React.render was removed in version 15 - probably want to update the examples in this PR too.

CMTegner: Yep, going to do it in a separate PR.

isnifer:
Hey, @sprjr! Nice to see that PR merged into master!

Issue #86, opened 2016-03-29
Isomorphic rendering broken


thebigredgeek:
https://github.com/reactjs/react-autocomplete/blob/master/lib/Autocomplete.js#L52 < This seems problematic...  Obviously the same ID won't exist on both sides of the pipe

danielfalk:
Can confirm as it happens to me.  Relevant line has shifted, so here's a more stable link:

https://github.com/reactjs/react-autocomplete/blob/6c7a7cfe15b161187a7c802eb0e19c257aaab20c/lib/Autocomplete.js#L52

danielfalk:
This is the same as the problem reference here: https://github.com/facebook/react/issues/5867

CMTegner:
This was fixed in #89. Should work fine in 1.0.0-rc4.

PR #95, opened 2016-04-11
Include an 'id' prop

danielfalk:
This is to address issue https://github.com/reactjs/react-autocomplete/issues/86

Add an 'id' prop to the component so that the id used for accessibility
can be managed externally if necessary.  At the moment, it is needed for
universal applications.

This prop is optional and will default to the old method of
autogenerating the identifier if one is not supplied.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -8,6 +8,7 @@ let Autocomplete = React.createClass({
 
   propTypes: {
     value: React.PropTypes.any,
+    id: React.PropTypes.string,
     onChange: React.PropTypes.func,
     onSelect: React.PropTypes.func,
     shouldItemRender: React.PropTypes.func,
@@ -55,7 +56,7 @@ let Autocomplete = React.createClass({
   },
 
   componentWillMount () {
-    this.id = lodash.uniqueId('autocomplete-'); 
+    this.id = this.props.id || lodash.uniqueId('autocomplete-'); 
     this._ignoreBlur = false
     this._performAutoCompleteOnUpdate = false
     this._performAutoCompleteOnKeyUp = false
--- a/examples/async-data/app.js
+++ b/examples/async-data/app.js
@@ -24,6 +24,11 @@ let App = React.createClass({
           attempt to autocomplete the first one.
         </p>
 
+        {/*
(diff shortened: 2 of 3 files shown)


lib/Autocomplete.js
```diff
@@ -55,7 +56,7 @@ let Autocomplete = React.createClass({
   },
 
   componentWillMount () {
-    this.id = lodash.uniqueId('autocomplete-'); 
+    this.id = this.props.id || lodash.uniqueId('autocomplete-');
```
bfink13: lodash is removed as part of #89

bfink13:
For someone who hasn't done much isomorphic rendering before, can you give a bit more explanation about why an id is necessary?

danielfalk:
Sure.  In isomorphic rendering, you first render your components on the server as a string and serve this to the browser, allowing the page to appear immediately rendered.  Once your application code is fully started up on the client side, the first thing React does is re-render the components and verify that the resulting dom tree is an **exact ** match of what the server sent.  If not, react gives us a nasty console message and has to throw away the dom and re-render the whole thing.

React however does a really good job of ensuring this works in every possible case.  Unfortunately there are still some cases which it cannot account for.  Introducing random id's into the resulting dom is one of those cases.  If the random id generated on the server happens to be different from that on the client, we will have a mismatch.

The answer I've proposed here is to allow the application that uses react-autocomplete to specify an id, so that the component can reliably operate in an isomorphic setting.

bfink13:
Got it, thanks for the context. Does the need for this go away with #89 (not generating a random ID) and the fact that the consumer can now set the ID on the wrapper and input (with inputProps and wrapperProps)?

lib/Autocomplete.js
danielfalk: https://github.com/reactjs/react-autocomplete/pull/89 appears to eliminate the label and `labelText` prop, which eliminates the need for a random id.  If that gets merged as it stands now, then this change is probably not necessary anyway.

danielfalk:
Yes.  The need goes away.

lib/Autocomplete.js
CMTegner: After #89 is merged you'll still have the ability to assign an id to the input via `inputProps` (which you should to make labels focus the corresponding input).

PR #102, opened 2016-05-09
Add default z-index to menu style

willclarktech:
Otherwise horrible-looking conflicts occur.


tiemevanveen:
I don't think this is a good idea, what if a user has a sticky menu with z-index of 1 or 2 and want to use this component somewhere on his page. With this z-index of 10, the autocomplete will "float" above the menu, that's now what you want right?

CMTegner:
`z-indez` is very situational, and whatever we pick as a default will probably be wrong for a lot of users. Closing since you can already easily specify `z-index` whenever you need to.

PR #105, opened 2016-05-13
Enable adding label properties

NunoPinheiro:
I currently need to style the label and this seemed like a good approach.


CMTegner:
[The label is now gone](https://github.com/reactjs/react-autocomplete/pull/89)! :)

PR #116, opened 2016-06-08
Tweaks

oliger:
Hi guys,

Here are some changes we've done to make the `react-autocomplete` works in one of our project.

Tell me what you think about those and if you want me to change stuff

Best,
Jimmy

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -310,7 +310,7 @@ let Autocomplete = React.createClass({
   },
 
   isInputFocused () {
-    var el = React.findDOMNode(this.refs.input)
+    const el = this.refs.input
     return el.ownerDocument && (el === el.ownerDocument.activeElement)
   },
 
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -6,7 +6,9 @@ let _debugStates = []
 let Autocomplete = React.createClass({
   propTypes: {
     value: React.PropTypes.any,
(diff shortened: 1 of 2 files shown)


blackrabbit99:
@oliger It's nice job but you've made several different fixes in one pull request. Could you please reduce it in different PR to make review easier?

oliger:
@blackrabbit99 Ok, will try to do this before the end of the week.

CMTegner:
@oliger just out of curiosity, what do you use `onFocus` and `onBlur` for?

oliger:
@CMTegner 

I need to be sure that the value in the fields are alway valid.

1. User start to type something
2. There is no autocomplete result so input value is invalid
3. User click outside the field
4. `onBlur` is triggered, set input value to what it was before user started editing it

Does that make sense to you?

CMTegner:
@oliger yeah, that sounds reasonable. I don't think it leads to great UX, though. I would get annoyed if I typed something into an input, only to have it cleared/reset because I made a typo. Perhaps highlighting the input as invalid would be better, which could be achieved via `onChange` (and `onBlur` if you want a delayed effect).

oliger:
> I don't think it leads to great UX

Yeah, but it was needed in the app I am working on.

I am also using `onBlur` / `onFocus` for tracking too.

CMTegner:
Closing since you can now specify any event handler in `props.inputProps`.

PR #119, opened 2016-06-16
Fix usage findDOMnode(deprecated)

### LINKED ISSUE #117: Warning for React.findDOMNode deprecated
Can this be fixed? it is just a easy fix.

blackrabbit99:
Fixed #117  That can be used into minor release


CMTegner:
I'm tempted to say just remove it entirely, since we only really support 0.14 and 15. AFAIK this API has been deprecated since 0.14.

blackrabbit99:
I see, totally agree with your suggestion

PR #124, opened 2016-06-24
Fixed "Uncaught TypeError: React.findDOMNode is not a function"

donbeave:


CMTegner:
This has already been fixed and pushed to npm in 1.0.1. Thanks for taking the time to make a PR!

PR #125, opened 2016-06-26
Remove deprecated React API usage (fixes #119)

CMTegner:


CMTegner:
@sprjr this one's pretty important, as react-autocomplete v1 doesn't work at all with React v15.

sprjr:
@CMTegner thanks! Agree on your other note we should probably start a `1.1.x` thread while we keep doing the patch versions when stuff comes up. I just found out I can't quite use this the way I'd like because auto complete does the matching for you explicitly (e.g. looks for the `getItemValue()` string at the beginning of  the input value) and was considering a custom `getMatchValue()` so users could optionally perform that however they'd like. I'm sure there's other things filed already as well...

CMTegner:
>auto complete does the matching for you explicitly (e.g. looks for the getItemValue() string at the beginning of the input value)

Yeah, this has been on my list of todos for a while. I've been meaning to create tickets for everything that needs attention, but I think we first need to do a proper triage of all the open tickets and PRs. At the moment there's a lot of noise in the form of duplicates, issues that are no longer valid, things we won't be fixing, etc.

PR #126, opened 2016-06-29
Fix: state.value to props.value

alvinsj:


CMTegner:
Thanks! Could you please move the second commit to a separate PR? We try to keep PRs focused on a single change. You may also want to check out the [contributing guidelines](https://github.com/reactjs/react-autocomplete/blob/master/CONTRIBUTING.md) on how to format the commit messages (we auto-generate changelogs based on commits).

CMTegner:
Fixed in 2fbafa64307e657d8ee37f71dd7560dde69e4637.

PR #132, opened 2016-07-09
[added] Ability to return custom components from renderMenu/renderItem (fixes #127)

### LINKED ISSUE #127: TypeError: doc is undefined (Firefox) / Uncaught TypeError: Cannot read property 'body' of undefined (Chrome)
I have setup an Autocomplete example based on the async one.
Results are loaded correctly, items are rendered correctly. But as soon as I **hover** an item I am getting an error.

Here the stack trace from Chrome (as Firefox stack trace still suck this days)

screen shot 2016-06-29 at 11 08 06

And here my component (wip)

```js
// @flow

import React, { Component } from "react"
import { StyleSheet, View } from "react-native"
import AutocompleteWeb from "react-autocomplete"

import fetchAPI from "~/modules/fetchAPI"

type Props = ReactJsonschemaFormWidgetProps & {
  // url to fetch the data
  url: string,
  formatItem: (unformattedItem: Object) => Item,
}
type State = {
  loading: boolean,
  items: Array<Item>,
  selectedItem: Item | null,
}

export type Item = {
  label: string,
  value: string,
}

export default
class Autocomplete extends Component<void, Props, State> {

  props: Props;

  state: State = {
    loading: false,
    items: [],
    selectedItem: null,
  };

  componentWillMount() {
    this.fetch()
  }

  getItemValue: Function = (item: Item): string => item.label;

  fetch: Function = (): Promise => {
    return (
      fetchAPI(this.props.url)
      .then((response) => {
        this.setState({
          loading: false,
          items: response.json.map(this.props.formatItem),
        })
      })
    )
  }

  handleSelect: Function = (value, item) => {
    // set the menu to only the selected item
    this.setState({
      selectedItem: item,
    })

    this.props.onChange(value)
  }

  handleChange: (event: SyntheticEvent, value: string) => void =
  (event: SyntheticEvent, value: string): void => {
    this.setState({
      loading: true,
    })
    this.fetch(value)
  };

  renderItem(item: Item, isHighlighted: boolean): ReactElement {
    return (
      <View
        id={ item.value }
        key={ item.value }
        style={ isHighlighted ? styles.itemHighlighted : styles.item }
      >
        { item.label }
      </View>
    )
  }

  render(): ReactElement {
    const { props } = this
    return (
      <AutocompleteWeb
        inputProps={ {
          style: styles.input,
        } }
        value={ props.value }
        required={ props.required }
        items={ this.state.items }
        getItemValue={ this.getItemValue }
        onSelect={ this.handleSelect }
        onChange={ this.handleChange }
        renderItem={ this.renderItem }
      />
    )
  }
}

const styles = StyleSheet.create({
  input: {
    borderWidth: 1,
    borderColor: "red",
  },
  item: {

  },
  itemHighlighted: {

  },
})

```

Any idea?

CMTegner:
Introduces a new peerDependency: react-dom

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -1,4 +1,5 @@
 const React = require('react')
+const { findDOMNode } = require('react-dom')
 const scrollIntoView = require('dom-scroll-into-view')
 
 let _debugStates = []
@@ -79,7 +80,11 @@ let Autocomplete = React.createClass({
     if (this.state.isOpen === true && this.state.highlightedIndex !== null) {
       var itemNode = this.refs[`item-${this.state.highlightedIndex}`]
       var menuNode = this.refs.menu
-      scrollIntoView(itemNode, menuNode, { onlyScrollIfNeeded: true })
+      scrollIntoView(
+        findDOMNode(itemNode),
+        findDOMNode(menuNode),
+        { onlyScrollIfNeeded: true }
+      )
     }
   },
 
--- a/lib/__tests__/Autocomplete-test.js
+++ b/lib/__tests__/Autocomplete-test.js
@@ -294,4 +294,27 @@ describe('Autocomplete#renderMenu', () => {
 
   });
 
+  it('should allow using custom components', () => {
+    class Menu extends React.Component {
+      render () {
+        return <div>{this.props.items}</div>
+      }
+    }
+    class Item extends React.Component {
+      render () {
+        return <div>{this.props.item.name}</div>
+      }
+    }
+    const wrapper = mount(AutocompleteComponentJSX({
+      renderMenu (items) {
+        return <Menu items={items} />
+      },
+      renderItem (item) {
+        return <Item key={item.abbr} item={item} />
+      }
+    }));
+    wrapper.setState({ isOpen: true, highlightedIndex: 0 });
+    expect(wrapper.find(Menu).length).to.equal(1)
+    expect(wrapper.find(Item).length).to.equal(50)
+  });
 });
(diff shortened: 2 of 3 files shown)


sprjr:
@CMTegner hey, I just noticed that I'm able to give you write permissions and all that kind of stuff here. I have some time planned this weekend for OSS stuff and if you'd like we can work on cleaning up the Issues / PRs, and perhaps get the next roadmap put together?

This particular PR looks good to me though - so we can merge it in whenever (either before or after a roadmap). Let me know!

CMTegner:
@sprjr I saw, thanks! I've already started going through some PRs and issues, but I'll have some more time tomorrow to do a proper sweep.

Perhaps we should start by closing items that haven't been touched in the last 6 months, then triage the remaining and see what we want to add to the 1.1/2.0 roadmap. How do we keep synced while doing this?

sprjr:
@CMTegner Weekend got away from me. Derp.

It might be a good idea to add things to a milestone that we'd like to tackle. I made one for `Release 1.1.0` for our next feature set, and `Release 1.0.x` for the next planned bugfix/patch releases. We could curate the issues into those milestones, and `wontfix` or close anything that we don't see being able to get out in the next few iterations?

It's great to have plans available on GH in general, although I have been wishing there were DMs on github some times so we could just _chat_ about some of this instead of the reply ping-pong.

Open to suggestions!

CMTegner:
@spicyj heh, same :p 

>We could curate the issues into those milestones, and wontfix or close anything that we don't see being able to get out in the next few iterations?

Sounds good to me, I was thinking along the same lines. I think it's important for our own sanity to clear out the old issues so we can get a better picture of what we should be focusing on, and also be able to use GH issues to plan the short/mid term roadmap.

I'll start adding to the milestones.

PR #135, opened 2016-07-11
[changed] set the position of the menu as fixed

coreoas:
This is a copy of the changes from https://github.com/reactjs/react-autocomplete/pull/75/files without the conflicts


CMTegner:
Just a heads-up (in case you're waiting for this PR to land): You can fully customize the dropdown menu by using `props.menuStyle` and `props.renderMenu`.

We'll probably be removing most of the default menu styles for 2.0, possibly only keeping the absolute coordinates which are passed to `props.renderMenu` (although you could calculate this yourself if we added support for rendering custom components instead of just fragments for the menu).

PR #137, opened 2016-07-18
explicitly set displayName on Autocomplete component. closes #2

### LINKED ISSUE #2: displayName on combobox
combobox component is missing its displayName (making quick notes as I start to bring this in)

sprjr:


CMTegner:
This is already handled by babel: https://github.com/reactjs/react-autocomplete/blob/master/build/lib/Autocomplete.js#L11

It's part of babel-core v5, and available through `babel-plugin-transform-react-display-name` / `babel-preset-react` in babel v6.

sprjr:
Ah, makes sense. I don't really like magic like that usually but since that's in the project I'm ok with leaving it out. I'll close #2 now then.

PR #139, opened 2016-07-18
[fixed] Reset `highlightedIndex` when it's outside of `items.length`

CMTegner:
I believe this will resolve most of the funkyness we've been seeing when updating `items` on the fly while the user types, e.g. #52. /cc @sprjr

### DIFF
--- a/lib/__tests__/Autocomplete-test.js
+++ b/lib/__tests__/Autocomplete-test.js
@@ -60,6 +60,21 @@ describe('Autocomplete acceptance tests', () => {
 
   });
 
+  it('should reset `highlightedIndex` when `items` changes', () => {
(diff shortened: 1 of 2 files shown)


lib/__tests__/Autocomplete-test.js
```diff
@@ -60,6 +60,21 @@ describe('Autocomplete acceptance tests', () => {
 
   });
 
+  it('should reset `highlightedIndex` when `items` changes', () => {
+    autocompleteWrapper.setState({ highlightedIndex: 10 });
+    autocompleteWrapper.setProps({ items: [] })
```
sprjr: nitpick: `;`

CMTegner: Fixed!

PR #146, opened 2016-07-21
[added] isAutoSelect prop.

alxpy:
These changes are necessary for the case, when I don't need to automatically select the first element.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -125,6 +125,9 @@ let Autocomplete = React.createClass({
   handleChange (event) {
     this._performAutoCompleteOnKeyUp = true
     this.props.onChange(event, event.target.value)
+    this.setState({
+      isAutoSelect: this.props.isAutoSelect
+    })
   },
 
   handleKeyUp () {
@@ -164,6 +167,7 @@ let Autocomplete = React.createClass({
       this._performAutoCompleteOnKeyUp = true
       this.setState({
         highlightedIndex: index,
+        isAutoSelect: true,
         isOpen: true,
       })
     },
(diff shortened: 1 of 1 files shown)


ddenis-ua:
+ I need this feature so much

CMTegner:
>These changes are necessary for the case, when I don't need to automatically select the first element.

Do you mean highlighting the top item in the menu, or the typeahead that appears in the input field?

alxpy:
@CMTegner, I mean both.

alxpy:
@CMTegner so what about my pull request?

CMTegner:
We're in the process of (temporarily) [removing the typeahead](https://github.com/reactjs/react-autocomplete/pull/152) functionality, since it's causing problems that will take a while to address. 

I'm not entirely sure why you want to disable the automatic highlighting of the first matching item. Could you provide a use-case where this might be useful?

alxpy:
It's important for UX of our website. I use react-autocomplete in search form. When user writes a search query and presses ENTER, he/she expects to see all the search results. But now we automatically select the first item and the user sees the result in a specific category.

Some case:
correctly: search query - `ip` (in all categories)
wrong:  search query - `iphone` (in specific category)
![search form without isAutoSelect]
(http://storage6.static.itmages.com/i/16/0801/h_1470051629_1701209_ae6e69d54f.png)

CMTegner:
@alxpy: Just to see if I'm understanding this correctly: You want `Autocomplete` to suggest items (by displaying `props.items`), but you want to prevent the top match from being highlighted so that if the user hits return (Enter) `props.onSelect` won't be triggered?

If yes: #159 solves this by touching as few code paths as possible.

alxpy:
Initially, I talked a bit about the other. But now it suits me.
Thank you!

Issue #148, opened 2016-07-23
Documentation for 1.x outline


jmfurlott:
It is true that the `PropTypes` and examples are more than enough for most users to get started, but is there any initiative to provide documentation?

I would be happy to help out but I wanted to ask some of the maintainers if they have an outline or anything started with sections that are more bitesized for contributing to.

CMTegner:
Argh, you got us! We were so happy pretending that "read the code" was an excuse for missing documentation! :p I've added this to the 1.1.0 milestone, since I've been meaning to create this issue for a while.

If you feel like helping out then a good starting point would be to simply list all the propTypes (I think they're all there, but best to double check the code) and provide a short explanation. I think the first version of the docs could just be put in the README.

CMTegner:
The readme has seen a lot of love lately. AFAIK this issue is no longer applicable. Just reopen if I'm wrong!

PR #149, opened 2016-07-27
Listing of Props in README to being basis of documentation

jmfurlott:
_An incomplete PR_

Just wanted to see if this was right direction -- I listed the props based on the `Autocomplete.js` file and wrote their PropTypes and a little bit about their functionality.  Is this what you were thinking about documentation? Wanted to ask before we go down the wrong path.

See https://github.com/reactjs/react-autocomplete/issues/148

Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,24 +28,42 @@ Check your work:
 
 # Props
 
-Because `react-autocomplete` is just one component, the main customization come
-from passing props to the `Autocomplete` component.
+Because `react-autocomplete` is just one component, the main customization comes
+from just passing props to the `Autocomplete` component. `react-autocomplete`
+will at least function with just the required props having been passed.
 
 ## Required props
 
+### `items` (array)
+
+This is an array of the items, themselves, to be put in the list for
+autocomplete to show as possible options and to compare against when the user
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,3 +25,45 @@ Write them:
...
+but let the user find values based solely on a specifically key.
+
+### `renderItem` (function)
+
+A function that returns a React element to show as an item in the list. Two
+props are passed to the function, `item` and `isHighlighted`, which allow the
+items in the list to be customized and styled. To style the container of these
+items, review the `menuStyle` prop.
+
+## Optional props
```
sprjr: @jmfurlott Required props looks great! maybe the additional props here would be better formatted with a simple table? [react-select](https://github.com/JedWatson/react-select#further-options) has a good example I think.

jmfurlott: Great idea and example.  Will have that for additional props next!

sprjr: Sure thing. Thanks for the help! Let me know when you think this PR is in a good spot and we can get it merged in.

CMTegner:
We now have documentation the the README. I'm sure this can be improved upon if you're still keen on working on the documentation.

Issue #111, opened 2016-05-30
Make text selection configurable


barinbritva:
Autocomplete is fully not usable on mobile devices. Entering a new symbol clears selection, but not adds letter. Need to type letter twice.

![ezgif com-video-to-gif 1](https://cloud.githubusercontent.com/assets/4758362/15648700/efa94878-2674-11e6-9b6c-91ed43982a7b.gif)

On some devices after second typing, letter duplicated.

barinbritva:
Something like this: #112

CMTegner:
I'd like to figure out why this is happening instead of adding a band-aid. Could you provide any more info about the environment you're seeing this problem in? I think the version running on reactcommunity.org is slightly outdated, so it would be great if you could provide an example with the latest version.

barinbritva:
It reproduces on version 1.0.0-rc5. All devices: ios, android, windows phone.

CMTegner:
I wonder if perhaps the native autocomplete is getting in the way (#93). Could you test the latest version from git and see if you can still reproduce?

meetmatt:
@CMTegner I was able to reproduce the issue as well and `autoComplete: 'off'` makes no difference.
+1 for configuration flag to turn it off completely, because in current state it makes it nearly unusable for mobile users.

sprjr:
Since this is making it entirely unusable for mobile users (or a sizeable % of devices for sure) I added this to the next patch release, and I think it'd be great if we could include a proper fix for this in that milestone.

@CMTegner I agree that the fix we do find would be best served to not be a bandaid if possible.

ImanMh:
Is it ready yet?

sprjr:
This should be resolved via #152, we'll put the selection typeahead on a feature enhancement and see when we can get that added back in while working for mobile as well.

Will try to have the patch release out soon (a day or two at latest, if not today)

ImanMh:
I downloaded v0.1.4 but it still highlights the ahead text. Do I have to change any of options?

CMTegner:
The latest official release is [1.1.0 on npm](https://www.npmjs.com/package/react-autocomplete).

ImanMh:
@CMTegner v1.1.0 did not work for me. Do I have to change any of my own code or downloading new version is enough?

CMTegner:
[...]

PR #151, opened 2016-07-27
[fixed] #111 removing the setSelectionRange functionality

sprjr:
This should resolve the issue mobile users are seeing. I don't think we even need this to be honest - especially if it blocks usability on an entire series of platforms. We highlight the current item, and you can still interact with the keyboard/enter to select an item - this selection range is a 'nice to have' when I look at it against issues like #111.

I think we should remove it, and then add it back in if it's worth the effort for more platforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+Unreleased (add items here not officially published)
+--------------------------------------
+- Remove `setSelectionRange`, preventing mobile users from being able to use the input
+
+
 v1.0.1 - 26 June 2016
 --------------------------------------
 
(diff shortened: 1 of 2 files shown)


sprjr:
I can't run tests locally (which I'll look in to), and get the broken tests fixed up so this build is passing.

CMTegner:
Mind waiting a day with this? I'm working on a change that makes this functionality optional, but I discovered some funky behaviour which I think may be caused by a timing/sequencing bug in the existing code. I want to get to the bottom of it so we don't accidentally release something that starts causing fires.

sprjr:
No worries! I mostly rushed it so we could unblock mobile - but fixing the deeper bugs would definitely be the better route. We can fall back to this as a worst case. I'll fix the tests in a bit (saw your comment on the other thread, I'll check in to that local issue) and then leave this branch be while you dig in to the other issue you're seeing here.

CMTegner:
Superseded by #152 (detailed explanation in description and comments).

Issue #80, opened 2016-03-07
Problems with cursor position in Internet Explorer


shiitake:
When you are typing in the input field if you match an item by more than a few characters it will narrow the search results and highlight the top option.  Once this happens if you try to type additional letters the cursor will have moved to the first position of the text field. 

This can be demonstrated using the examples linked on the site: 
https://reactcommunity.org/react-autocomplete/async-data/

CMTegner:
Reproducible in IE11 on Windows 8.1:

1. Navigate to https://reactcommunity.org/react-autocomplete/async-data/
2. Type 'A' in the input
3. Observe that the 'Alabama' is highlighted in the autocomplete menu
4. Type 'L' in the input
5. Observe that the input value is 'LA', expected: 'AL'

I'm not sure what causes this, but if no one else has time I'll try to have a look this weekend.

HenriJ:
Hi @CMTegner . Did you have any luck finding the root cause ? We are stuck because of this problem too, and can have a look also if it can help.

CMTegner:
I've had little look at the problem. My initial suspicion was that `input#setSelectionRange()` (which is invoked when Autocomplete fills the input with the top result) was bugging out in IE, and that's sort of what's happening. To exclude the possibility of `setSelectionRange()` being completely borked on IE I made a quick test page:

```html
<!doctype html>
<html>
<head>
  <title>setSelectionRange test</title>
</head>
<body>
  <input type="text" value="foobar">
  <script type="text/javascript">
    var input = document.querySelector('input');
    input.focus();
    input.setSelectionRange(1, 6);
  </script>
</body>
</html>
```

Load that up in IE11, type a character, success.

After having played around with `setSelectionRange` for a while, I came to the conclusion that if you create a range with a non-zero length, the caret (or at least the internal caret state position) is moved to position 0 (the start of the input). In other words, if you use `setSelectionRange` to place the caret (end = start) after updating the input value, things work as expected. The problem occurs when you create an actual text selection range (end > start).

My current theory is that Autocomplete's usage of `setSelectionRange()` is somehow interfering with React's focus/caret positioning logic. I haven't had time to dig any deeper.

HenriJ:
Not sure exactly why it behaves like that (as you say, it might be interferences between React and setSelectionRange), but I found a workaround that seems to fix the issue : forcing the position of the caret in the callback of the setState() that was causing the problem.

[...]

PR #152, opened 2016-07-28
[removed] Remove typeahead (fixes #40, #111)

### LINKED ISSUE #40: Android: Delete doesn't work as expected
#Nexus 5, delete issue

It's hard to show this issue but I have captured it via USB debugging. When you try and delete the last letter it doesn't change the value, then (seemingly) randomly after a certain amount of attempts it will delete the whole field.

You can see on the right hand side when the elements flash I am typing in the box.

react-autocomplete

CMTegner:
I initially started by trying to fix a regression to #80 which was introduced
in 2fbafa, but after having evaluated the code I discovered that there's no
easy way to both fix the regression and make the typeahead logic work again.
This is sadly a side-effect of no longer keeping `value` in state, since we
can no longer diff previous state and next state in the same fashion.

Since the current typeahead situation is very poor on mobile, not to mention
how broken it is in IE, I decided the best way forward is to temporarily remove
the typeahead logic.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -103,18 +103,9 @@ let Autocomplete = React.createClass({
     if (this.keyDownHandlers[event.key])
       this.keyDownHandlers[event.key].call(this, event)
     else {
-      const { selectionStart, value } = event.target
-      if (value === this.props.value)
-        // Nothing changed, no need to do anything. This also prevents
-        // our workaround below from nuking user-made selections
-        return
       this.setState({
         highlightedIndex: null,
         isOpen: true
-      }, () => {
-        // Restore caret position before autocompletion process
-        // to work around a setSelectionRange bug in IE (#80)
-        this.refs.input.selectionStart = selectionStart
       })
     }
   },
@@ -235,17 +226,8 @@ let Autocomplete = React.createClass({
     var itemValueDoesMatch = (itemValue.toLowerCase().indexOf(
       this.props.value.toLowerCase()
     ) === 0)
-    if (itemValueDoesMatch) {
-      var node = this.refs.input
-      var setSelection = () => {
-        node.value = itemValue
-        node.setSelectionRange(this.props.value.length, itemValue.length)
-      }
-      if (highlightedIndex === null)
-        this.setState({ highlightedIndex: 0 }, setSelection)
-      else
-        setSelection()
-    }
+    if (itemValueDoesMatch && highlightedIndex === null)
+      this.setState({ highlightedIndex: 0 })
   },
 
   setMenuPositions () {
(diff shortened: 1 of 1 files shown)


lib/Autocomplete.js
```diff
@@ -103,18 +103,9 @@ let Autocomplete = React.createClass({
     if (this.keyDownHandlers[event.key])
       this.keyDownHandlers[event.key].call(this, event)
     else {
-      const { selectionStart, value } = event.target
-      if (value === this.props.value)
```
CMTegner: This was previously `value === this.state.value`, which caused the next block of code to execute every time, and breaking the workaround which made user-defined selections not break. When I "fixed" it by changing it to it's current form I overlooked the fact that `event.target.value` is always equal to `this.props.value`, since React prevents the input's value from ever changing.

lib/Autocomplete.js
```diff
@@ -103,18 +103,9 @@ let Autocomplete = React.createClass({
...
-        // Nothing changed, no need to do anything. This also prevents
-        // our workaround below from nuking user-made selections
-        return
       this.setState({
         highlightedIndex: null,
         isOpen: true
-      }, () => {
-        // Restore caret position before autocompletion process
-        // to work around a setSelectionRange bug in IE (#80)
-        this.refs.input.selectionStart = selectionStart
```
CMTegner: This line of code without the fix above causes user-made selection (i.e. shift+arrow) to be completely broken. And this, to reiterate, was the fix for the incredibly weird typeahead bug we had in IE. Without this fix `Autocomplete` is unusable in IE.

CMTegner:
I'll update the CHANGELOG and README (to mention that we currently don't have typeahead) once this is merged. I noticed we're missing a few entries in the former. Oh, and during my bug hunt I discovered a substantial amount of code that can be removed/simplified now that `value` only changes when passed in by the consumer. I'll get to work on that this weekend.

CMTegner:
Just to be crystal clear: I really do want to figure out how to make typeahead work across platforms and devices, because I think it's a really cool feature. I'm not sure in which form though, because after som a11y testing I discovered that the current implementation which uses text selection ranges is...not great if you're using a screen reader.

sprjr:
Heya, sorry for the radio silence there - and for chasing the yaks on my build PR/comments. This looks good to me, and I think this gets us pretty close on the next release then if I'm not mistaken.

Anything else new that should go in the next minor/patch (need to look in to whats there. I'll open a PR with the changelog/version proposal here shortly if the milestone looks pretty clear.

CMTegner:
> Heya, sorry for the radio silence there - and for chasing the yaks on my build PR/comments.

No worries; here to help! :)

PR #154, opened 2016-08-01
[fixed] ie menu closing issue

kortv:
Menu does not closing after selecting one. ie onliy

When you try to select one item from the menu, the menu is not closing. For me this is a problem on the IE11 in not IE browsers work fine.
This can be demonstrated in ie11 using the examples linked on the site: 
https://reactcommunity.org/react-autocomplete/async-data/

this is happens because that, in the method selectItemFromMouse  this.refs.input.focus() occurs after this.setIgnoreBlur(false). I found for me a simple solution to adding all setTimeout other browsers works still well.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -274,7 +274,9 @@ let Autocomplete = React.createClass({
     }, () => {
       this.props.onSelect(value, item)
       this.refs.input.focus()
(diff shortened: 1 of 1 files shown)


CMTegner:
Superseded by 61a4e05038c4fd79a6dda6dfbc6169bfeed3a22d.

PR #156, opened 2016-08-02
[release] 1.1.0

sprjr:
Looks like there's been a log of activity since `1.0.1` on master. As some of the things felt more like minor updates (mostly the custom menu  & examples) when reviewing the commits I thought `1.1.0` might be more appropriate. If we want to cut a `1.0.x` branch and perhaps remove some of that with only the bug fixes we could as well - but I'm not sure if folks really need that or would prefer it.

I also updated the CHANGELOG format, let me know if you think this is actually better - I've started using it recently and prefer it to other ad-hoc or generated ones.


CMTegner:
Yeah, some 1.1 things were merged prematurely, and I don't think we want to start cherry-picking a 1.0.2.

There are a few ways in which we can improve on this going forward:

1. Reserve master for bugfixes, and maintain a 1.x branch which we merge new features into, then merge with master when we release
2. Maintain a <latest release> branch which only gets bugfixes, while bugfixes and features buffer on master while we work on the next release
3. Cut releases more frequently so that the latest release is ~always up to speed with master

>I also updated the CHANGELOG format, let me know if you think this is actually better - I've started using it recently and prefer it to other ad-hoc or generated ones.

I'm a fan of the keepachangelog.com format, so 👍 from me on that.

sprjr:
I think 3 is something we can work on independent of 1 and 2. I don't have a preference strongly either way - although I like the idea of a `release` or `next` branch in which all minor/major type changes are merged, and reserving master for bugfixes so that it's 'clear for patch release' at almost any given point in time.

I can send a PR with that outline (unless you prefer another solution) in `CONTRIBUTING.md` if you'd like as well.

CMTegner:
Do it 👍  I think I prefer to keep master as the most up-to-date source of truth (i.e. option no.2). I'm happy to help out with releases too, if you feel comfortable with that.

sprjr:
@CMTegner that'd be good - I didn't realize til just now I could also grant access to that repo. Want to get me your npm user name here or on Twitter?

CMTegner:
@sprjr cmtegner, all lowercase :)

PR #159, opened 2016-08-07
[added] autoHighlight prop to toggle automatic highlighting of top match

CMTegner:
As requested by #146. Not sure about the prop name, so I would appreciate some feedback. I originally went with `highlightFirst`, but didn't feel that was good enough.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -19,6 +19,7 @@ let Autocomplete = React.createClass({
     inputProps: React.PropTypes.object,
     wrapperProps: React.PropTypes.object,
     wrapperStyle: React.PropTypes.object,
+    autoHighlight: React.PropTypes.bool,
     debug: React.PropTypes.bool,
   },
 
@@ -45,7 +46,8 @@ let Autocomplete = React.createClass({
         position: 'fixed',
         overflow: 'auto',
         maxHeight: '50%', // TODO: don't cheat, let it flow to the bottom
-      }
+      },
+      autoHighlight: true,
     }
   },
 
@@ -214,7 +216,7 @@ let Autocomplete = React.createClass({
   },
 
   maybeAutoCompleteText () {
-    if (this.props.value === '')
+    if (!this.props.autoHighlight || this.props.value === '')
       return
     var { highlightedIndex } = this.state
     var items = this.getFilteredItems()
--- a/lib/__tests__/Autocomplete-test.js
+++ b/lib/__tests__/Autocomplete-test.js
@@ -98,6 +98,22 @@ describe('Autocomplete keyPress-><character> event handlers', () => {
     done();
   });
 
+  it('should highlight top match', () => {
+    var autocompleteWrapper = mount(AutocompleteComponentJSX({}));
+    autocompleteWrapper.setState({ isOpen: true });
+    autocompleteWrapper.setProps({ value: 'a' });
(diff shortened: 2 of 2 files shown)


alxpy:
Prop name is good! When will release with this code?

sprjr:
I think `autoHighlight` works. I tried to think of something that might work as well, and all I could come up with is `useHighlight` which maybe only makes more sense from the code standpoint and not the prop.

I think we should merge it and 🚢 it. Did you get a chance to get me your npm handle for releases btw?

sprjr:
@CMTegner I added you to the npm access (per the other thread).

The only reason I did not go ahead and publish the new version is that I overlooked the `CHANGELOG.md` was not included in the PR and I wasn't quite sure what version area to bump here. If you want to add your stuff along with the current unreleased change (the babelrc changes already merged) and then publish this I think we're good to.

CMTegner:
@sprjr I updated the changelog, but I'm hesitant to publish since I don't really know the process. I'm assuming you're using `rackt release`, but as far as I can tell this makes changes to changelog, which then get published to npm before you're able to make edits.

How do you handle this?

sprjr:
@CMTegner actually, I didn't like the 'auto changelog if your commits are formatted right' behavior so I've just been releasing this manually. Usually I run the build, tests, then I publish it, then I commit the changes/changelog with a manual `[release] {version}` message. I can publish this out though if it's good to go otherwise - was more wondering what the new version number / changelog updates would be actually.

sprjr:
@CMTegner I went ahead and published 1.2.0, since it was a new prop I felt like that was more semantic. Let me know if you disagree moving forward.

Also - the `.babelrc` changes actually _broke_ my `rackt build` command. If I reverted the file, I was able to run it to generate the build (to the old `.env.test.presets` - so we'll have to figure out what's going on there at some point otherwise releases have this weird step...

I'll get the gh-pages up to date as well now.

CMTegner:
So you publish from inside `/build`, right? I started work on replacing `rackt-cli` last weekend. I think we'll benefit from having something that's a bit more flexible.

CMTegner:
I should probably elaborate a little: I noticed that `npm start` no longer works because babel v5 doesn't recognize `presets` at the top level in `.babelrc`. I'm getting pretty tired of having to deal with weird build issues due to - amongst others - mismatch between our local babel version and the one that `rackt-cli` is running. I don't really want to do the counter-intuitive thing and downgrade ours, so I'd rather replace the entire setup with something more simple:
* babel-cli to transpile the source
* browserify to create UMD bundles and build/run examples
* A simple, npm-run-script-powered deployment:
 - npm version <version>
 - npm run build
 - npm publish

CMTegner:
Mkay, apparently GitHub didn't want to show me your second comment until now. Apologies for the duplicate observations.

> I went ahead and published 1.2.0, since it was a new prop I felt like that was more semantic. Let me know if you disagree moving forward.

Totally agree, that was the same version I had planned to publish.

sprjr:
Hmm, ya sorry for the frustrations there. When I was trying to get `npm test` working I hadn't even tried the other commands.

I've typically thought the rackt cli was neat but only if you made it or used it for your components. If you're new to it, it's kinda proprietary. 

I'm all for removing it, although of late between the build and other issues I've only been able to comment on PRs or release things, but not actually dig in to the tasks. So I think you're the best one to make the call there on what would work for you and anyone else wanting to contribute.

Thanks for all the collaboration / effort on this.

Issue #129, opened 2016-07-08
Expose internal state/interesting changes to a parent via a hook


docccdev:
Can pass isOpen through the props?

CMTegner:
I don't entirely understand what you're asking. Do you wish to control `state.isOpen` via props? What would be the use-case for this?

docccdev:
Yes exactly. I need it for the parent component to know about the state of isOpen

docccdev:
Parent component changes its appearance on the state.isOpen of autocomplete.

CMTegner:
So something like `props.onMenuVisibilityChange`?

```js
render () {
  return (
    <Autocomplete
      onMenuVisibilityChange={isOpen => this.setState({ isAutocompleteOpen: isOpen })}
    />
  )
}
```

docccdev:
Excellent solution is right for me :)

sprjr:
Given the discussions around #138, it might be better to generalize this, e.g. `onUpdate` or something like that, which would then invoke a single callback and pass a map of the current state values that a composer may wish to know?

CMTegner:
Perhaps. I wonder if there's a common way to expose internal state. Do you know of any other reusable components that do something similar?

westongame:
Hi. Can you tell me pls, in witch version are you planning to implement this feature?

CMTegner:
Closing this since the original request is covered by #163 (published as v1.3).

PR #163, opened 2016-08-31
Proposal: Managed menu visibility

CMTegner:
This proposal adds two new props, `onMenuVisibilityChange: func` and `open: bool`. The former is invoked every time `state.isOpen` changes, and receives said value as its only argument. The latter - when present - overrides `state.isOpen` by putting the component into "managed menu visibility mode" (which is just a fancy way of saying `props.open` takes precedence over `state.isOpen`).

I created a new example (examples/managed-menu-visibility) to demonstrate the flexibility these additions offer.

This covers the request made in #129.

I personally need this feature to be able to use Autocomplete in a situation where I want to force the user to pick an option from the menu. Having the menu close when the input is blurred is needless to say not ideal ;)

@sprjr I'd like your input before merging.

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -8,4 +8,5 @@
     <li><a href="static-data/">Static Data</a></li>
     <li><a href="async-data/">Async Data</a></li>
     <li><a href="custom-menu/">Custom Menu</a></li>
+    <li><a href="managed-menu-visibility/">Managed Menu Visibility</a></li>
   </ul>
--- a/examples/managed-menu-visibility/app.css
+++ b/examples/managed-menu-visibility/app.css
@@ -1,0 +1,21 @@
+body {
+  color: #333;
+  font-family: "Helvetica Neue", Arial, sans-serif;
+  font-weight: 200;
+}
+
+.example {
+  padding: 0 25px;
+}
+
+label {
+  display: block;
+  margin: 5px 0;
+}
+
+code {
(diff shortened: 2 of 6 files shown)


sprjr:
This looks good from the description of the design, and I agree being able to manage something is quite nice. For exposing the internal state, we'd talked about doing that for perhaps a few props - is that satisfied by this for now as its really the only state we want to peer at?

I'll check the code out here shortly when not en route to work.

CMTegner:
@sprjr good to have you back! How was react rally?

I decided to isolate the one internal attribute that was needed for this issue (`isOpen`) instead of having to come up with a clever solution for surfacing all internal state. I'm not sure how much more state we would want to make public, apart from perhaps `highlightedIndex` (which you already have access to in `renderItem`).

My thinking was that we could add this, then deprecate it if/when we decide on a more generic approach.

sprjr:
React Rally was great! A lot of really great talks, and an overall really great community / event. The implementation of this looks simple & clean enough to me - and covers the necessary use cases.

🚢  :shipit:!

CMTegner:
Published in 1.3.0. I had to follow up with 1.3.1 because for some reason `npm pack` included both `node_modules` and `coverage`, even though they're listed in `.gitignore`. Think I'll add `files` to `package.json` so this doesn't happen again.

PR #123, opened 2016-06-20
added onBlur and onFocus props

erichhsun:
Updated version of pull request #79

PR #165, opened 2016-09-02
[fixed] Prevent event handler clobbering (fixes #123)

CMTegner:
This change allows passing any prop to the `<input>` element via `props.inputProps`, apart from the ones we currently don't want the consumer to change (`role`, `aria-autocomplete`, and `autoComplete`).

The actual change in this PR is to ensure any event handlers we apply are added in addition to whatever the consumer specifies, instead of overwriting them.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -338,10 +338,11 @@ let Autocomplete = React.createClass({
       })
     }
 
+    const { inputProps } = this.props
     return (
       <div style={{...this.props.wrapperStyle}} {...this.props.wrapperProps}>
         <input
-          {...this.props.inputProps}
+          {...inputProps}
           role="combobox"
           aria-autocomplete="list"
           autoComplete="off"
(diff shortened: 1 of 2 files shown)


lib/Autocomplete.js
```diff
@@ -330,6 +330,12 @@ let Autocomplete = React.createClass({
     this._ignoreClick = false
   },
 
+  composeEventHandlers (internal, external) {
```
CMTegner: I could have done something fancy here, like
```js
composeEventHandlers (...handlers) {
  return e =>
    handlers
      .filter(v => v)
      .forEach(handler => handler(e))
}
```

...but I decided to optimise for the actual use case we have now, which is that the first is always present, while the second is unknown.

pcalves:
Looking forward for this to be merged. 🙌 And I'm probably refactoring some of the code on my current project to compose event handlers like you're doing here, really simple and elegant soluition.

CMTegner:
@sprjr no worries dude, we're all just finding time where we can ;)

These additions were published in 1.4.0. Go check them out!

PR #167, opened 2016-09-27
added onBlur and onFocus props

rogeriochaves:
This solves conflicts of #123


hanford:
I had to switch to react-autosuggest b/c of this problem, if you haven't checked it out .. I would. API / Mobile UX is much better @rogeriochaves

rogeriochaves:
@hanford I didn't knew that, thanks for the (auto)suggestion :)

CMTegner:
Hi! `onBlur` and `onFocus` - as well as any other event handler you might want -  are now supported via `props.inputProps`. Check out 1.4.0 and let us know if anything doesn't work as expected.

PR #173, opened 2016-10-27
Deprecate string ref

alvinsj:


CMTegner:
Hi! Any chance you could reduce this PR to only contain the changes to `ref` (minus the built files and all the whitespace changes)?

PR #184, opened 2016-11-28
Update package.json main to point to actual main file

Daniel15:
index.js simply does `module.exports = require('./Autocomplete')`, which is pointless. The package should instead point directly to Autocomplete.js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "react-autocomplete",
   "version": "1.4.0",
   "description": "Accessible, extensible, Autocomplete for React.js",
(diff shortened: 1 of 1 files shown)


CMTegner:
You're certainly not wrong! I'm going to hold off on this until we do the 2.0 API redesign. Ideally we want to remove `index.js` when we change `main`, but we can't in case people are requiring it directly (although it is of course undocumented).

Daniel15:
You can do this change without deleting the index.js file if you like :)

CMTegner:
/me notes down "remember to remove index.js"

PR #187, opened 2016-11-28
Add build instructions to readme

Daniel15:
This seems useful to know. 😛


natevw:
Thanks for this tip! +1 on this, although even better might be to integrate it into package.json as a script.

CMTegner:
Fixed in edb7774cf2cd2ba646d63459aaa8fffab0b7dc86.

PR #188, opened 2016-11-28
Nicer sorting algorithm for example

Daniel15:
The sorting in the example makes an effort to be clever, by sorting based on the location of the match. This is a great idea, but for the best user experience it should be complemented by alphabetical sorting when the match is in the same location (or there is no match).

For example, with the current version, searching for "e" shows the results in a confusing order:

After my change, all the states with "e" as their second letter are sorted alphabetically, then they're followed by other states with "e" elsewhere:

Similarly, with the empty state, the current sorting method is pretty hectic:

After my change, it's nice and alphabetical:

I also added a comment explaining what's going on, so people looking through the code understand why it's written this way 😛

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -24,11 +24,24 @@ export function matchStateToTerm (state, value) {
   )
 }
 
+/**
+ * An example of how to implement a relevancy-based sorting method. States are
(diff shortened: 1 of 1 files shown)


CMTegner:
Great, thanks!

PR #195, opened 2016-12-28
fix #176 - ownerDocument error ..for now

### LINKED ISSUE #176: Uncaught TypeError: Cannot read property 'ownerDocument' of null
This error occurs when user selects an item from the menu and then continues typing. I guess that using reference callback on menu object instead of string could solve this problem

tiemevanveen:
https://github.com/reactjs/react-autocomplete/issues/176 

> Uncaught TypeError: Cannot read property 'ownerDocument' of null #176

> This error occurs when user selects an item from the menu and then continues typing.

Seems itemNode is undefined in that case, this fixes that. 

Let's also fix https://github.com/reactjs/react-autocomplete/pull/173 so we get rid of those string refs

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -99,11 +99,13 @@ let Autocomplete = React.createClass({
     if (this.state.isOpen === true && this.state.highlightedIndex !== null) {
       var itemNode = this.refs[`item-${this.state.highlightedIndex}`]
       var menuNode = this.refs.menu
-      scrollIntoView(
-        findDOMNode(itemNode),
-        findDOMNode(menuNode),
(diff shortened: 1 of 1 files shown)


RubaXa:
Ping...

RubaXa:
@tiemevanveen Please, rebuild https://github.com/tiemevanveen/react-autocomplete/blob/fix-176/

Because
 - https://github.com/tiemevanveen/react-autocomplete/blob/fix-176/package.json#L5
 - https://github.com/tiemevanveen/react-autocomplete/blob/fix-176/build/lib/Autocomplete.js#L112

CMTegner:
Released in 1.4.2.

PR #196, opened 2016-12-28
Deprecate string ref

teameh:
Rebased and cleaned up version of https://github.com/reactjs/react-autocomplete/pull/173 

LGTM!

@alvinsj I hope you're alright with me cleaning up your commit.

### DIFF
--- a/examples/async-data/app.js
+++ b/examples/async-data/app.js
@@ -26,7 +26,7 @@ let App = React.createClass({
         <label htmlFor="states-autocomplete">Choose a state from the US</label>
         <Autocomplete
           inputProps={{name: "US state", id: "states-autocomplete"}}
-          ref="autocomplete"
+          ref={c => this.autocomplete = c}
           value={this.state.value}
           items={this.state.unitedStates}
           getItemValue={(item) => item.name}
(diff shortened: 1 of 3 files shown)


CMTegner:
Contradictory to what I wrote in #173, I don't think we can actually land this in 1.0. I know of several production apps that would break if we rename the (undocumented) refs. Currently refs are the only way of accessing the imperative API of the inner `<input>`.

I have [some ideas](https://github.com/reactjs/react-autocomplete/commit/6c44944f485af163906fc3aebe1f904858956364) on how to solve this for 2.0.

tiemevanveen:
>  I don't think we can actually land this in 1.0. I know of several production apps that would break if we rename the (undocumented) refs. 

Haha okay.. But then you can't make any breaking changes in 1.0?

> Currently refs are the only way of accessing the imperative API of the inner <input>.

Shouldn't that be included in v1.0 as well then?

tiemevanveen:
> I have some ideas on how to solve this for 2.0.

Ah read to fast, okay, yeah that might make sense, you could also skip v1.0 altogether.

CMTegner:
Fixed in f915315bfd89b6ab77b2125192b621cde609d1dc. We're still using the `this.refs` namespace for backwards compatibility.

PR #197, opened 2016-12-30
API documentation

CMTegner:
This change introduces a generated `README.md` which is built using `README-template.md` and scripts from `bin/`. At the moment the scripts only generate the API documentation and dumps it into the `$API_DOC` placeholder found in the template. The documentation is generated by using `react-docgen` to extract comments from the source code.

The README is generated by running `npm run generate-readme`. This can either be done in the PR which introduces a change, or just before publishing (as described in "Publishing / Releasing").

@sprjr: Mind giving it a once-over before I merge this?

### DIFF
--- a/bin/generate-api-doc.js
+++ b/bin/generate-api-doc.js
@@ -1,0 +1,78 @@
+'use strict'
+
+const fs = require('fs')
+const map = require('lodash.map')
+const flow = require('lodash.flow')
+const sortBy = require('lodash.sortby')
+const docgen = require('react-docgen')
+
+const TYPES = {
+  any: 'Any',
+  array: 'Array',
+  func: 'Function',
+  bool: 'Boolean',
+  string: 'String',
+  number: 'Number',
+  object: 'Object',
+}
+
+const SECRET_PROPS = [
+  'debug',
+]
+
+function parse (filePath) {
+  return docgen.parse(fs.readFileSync(filePath, 'utf-8')).props
+}
+
+function toArray (props) {
+  return map(props, (prop, name) => Object.assign({ name }, prop))
+}
+
+function prune (props) {
+  SECRET_PROPS.forEach(secret => {
+    if (!props.some(p => p.name === secret)) {
+      throw new Error(`\`${secret}\` is marked as a 'secret' prop, but it's not present in \`propTypes\``)
+    }
+  })
+  return props.filter(p => !SECRET_PROPS.includes(p.name))
+}
+
+function sort (props) {
+  return sortBy(props, [p => !p.required, 'name'])
+}
+
+function prepareDescription (props) {
+  return props.map(prop => {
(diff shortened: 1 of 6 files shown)


bin/generate-api-doc.js
```diff
@@ -0,0 +1,78 @@
+'use strict'
+
+const fs = require('fs')
+const map = require('lodash.map')
+const flow = require('lodash.flow')
+const sortBy = require('lodash.sortby')
```
sprjr: Not a big deal as its for development here, but I've typically opted for just `require('lodash/{module}')` - I know that can add a lot to the final module size if that's a concern but I believe there's other ways around that if it's a major concern as well.

lib/Autocomplete.js
```diff
@@ -7,20 +7,106 @@ let _debugStates = []
...
+     *
+     * Invoked when the user selects an item from the dropdown menu.
+     */
     onSelect: React.PropTypes.func,
+    /**
+     * Arguments: `item: Any, value: String`
+     *
+     * Invoked for each entry in `items` and its return value is used to
+     * determine whether or not it should be displayed in the dropdown menu.
+     * By default all items are always rendered.
```
sprjr: I take it this is to give more control to the passed items on if they display in the list or not? (e.g. when the input value is updated by the user?)

package.json
```diff
@@ -34,9 +35,13 @@
     "eslint": "^2.13.1",
     "eslint-config-rackt": "^1.1.1",
     "jest": "^13.0.0",
+    "lodash.flow": "^3.5.0",
+    "lodash.map": "^4.6.0",
+    "lodash.sortby": "^4.7.0",
```
sprjr: This is why I use `lodash/{module}` mentioned above, just so folks don't have to npm install or update specific versions too much

bin/generate-api-doc.js
CMTegner: Yeah, I'm not really fussed either way. The only real reason for going with individual dependencies is it makes it *slightly* easier to see which parts of lodash we're using, just by looking at `package.json`. This code is only for development, so size isn't really a factor.

package.json
CMTegner: I definitely see the convenience of having one import for development instead of multiple. I'll change it in a future PR.

lib/Autocomplete.js
CMTegner: Yes. There's no real reason to have `props.shouldItemRender` any longer. Now that `Autocomplete` is a managed component you need to keep track of `props.value`, so there's no reason why you can't just filter `props.items` before passing it in. Before, when we kept `value` in state, there was no guarantee that the consumer would be tracking `value`, so it made more sense to provide this convenience API so you could get more done "within" the component itself. I think we'll remove this prop for v2.

PR #200, opened 2017-01-11
Replace rackt-cli with custom build

CMTegner:
Preface: You may wonder why we would want to replace something that was seemingly working with a custom setup. The fact is that `rackt-cli` isn't actively maintained, so we suffered from being stuck with a very little flexible setup that had very few downstream changes. After this change lands well have full control over every aspect of building and releasing the module.

The new build system is fairly simple:
* `babel` is used to transpile the source, which becomes the output for the module which is published on `npm`
* `browserify` is used to create the UMD bundle and build and hot-reload the examples (during development)
* Everything is tied together with bash scripts for convenience

The main bulk of the diff is attributed to removing `dist/` and `build/` from the repo. These directories are only included in the published tarball. This a) keeps generated sources out of the repo and prevents us from accidentally forgetting to push the build result after a new version is published and b) removes the need for contributors to worry about whether or not to include generated sources in their PRs.

Some caveats with this change:
* ~~The UMD build no longer has source maps. This is due to the `prelude.js` hack needed to account for the external dependencies. `webpack` handles this better OoTB than `browserify` (if I have time I'll see if I can upstream a fix for this). The minified version still has source maps.~~ **This bothered me so much that I created [youemdee](https://www.npmjs.com/package/youemdee) to deal with the problem. We now have source maps for the UMD build without having to do any custom work in this repo.*
* Running the examples is pretty awkward since they have to be built and run individually (this is also something `webpack` does better). The gross details are hidden from the developer (`npm start` still works the same), but I'd like to overhaul the examples entirely. My goal is to create a proper homepage with examples and code side-by-side, and then use something else for development (perhaps [storybook](https://getstorybook.io/) or something similar).

Future:
* I'm still not sure if we should stick with `gh-pages` for the examples or switch to `docs/`. The former keeps a bunch of generated source out of the main repo, but the latter prevents us from having to deal with the awkwardness which is updating the `gh-pages` branch. I'm open to suggestions. This also relates to the above comment regarding a potential homepage.
* ~~It would be great to get some numbers on how popular the UMD bundle is. We could cut a lot of build cruft by removing it.~~ *No longer a problem.*


.babelrc
```diff
@@ -1,3 +1,8 @@
 {
-  "presets": ["es2015", "stage-0", "react"]
+  "presets": ["es2015", "stage-0", "react"],
```
sprjr: are we using `stage-0` still? might want to bump this up higher if we can get away with it

.babelrc
```diff
@@ -1,3 +1,8 @@
...
+  "env": {
+    "development": {
+      "presets": ["react-hmre"]
```
sprjr: not a huge deal, but (and I didn't know about this preset!) this is not maintained any more. Should be ok but is there something else we can swap it out with down the road or at least note that it's something that could be an issue later?

package.json
```diff
@@ -13,12 +18,17 @@
     "example": "examples"
   },
   "scripts": {
-    "generate-readme": "node bin/generate-readme.js",
-    "lint": "eslint lib examples bin",
-    "pretest": "npm run lint",
-    "test": "jest",
-    "coverage": "jest --coverage",
-    "start": "rackt server"
+    "generate-readme": "scripty",
```
sprjr: Never seen this module before, seems pretty ok if you're comfortable with it but it looks foreign to me to just have all the npm scripts point to a single `scripty` thing. Mostly just sounds like magic (which is kinda what rackt-cli was too, to me).

scripts/test/jest.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh -e
+
+jest
```
sprjr: I don't necessarily mind this set up if you really like it. Is there any value over 'consistency' here in setting `npm run jest` to point to `scripty` which effectively just runs `jest`? For some of the more complex tasks above, yea I can see the value. Am wondering if this will deter future collaboration or add confusion for folks (me) not familiar with this way of running the show.

.babelrc
CMTegner: Good point. I'll follow up with a PR to narrow down the list of transforms. We can probably get away with stage-3.

.babelrc
CMTegner: Yep, I'm aware of it. According to Dan Abramov a new version is on the horizon, but no news yet. If it starts causing problems we can just remove it and fall back to only using browserify-hmr (that'll give the same behaviour as we have today, without this change).

package.json
CMTegner: It's sort of a shorthand for pointing to individual shell scripts, with some added benefits like being able order multiple targets under one umbrella (e.g. "build", "build:component", "build:examples"). The only downside (as I see it) is that `npm run` becomes less helpful if you're not familiar with scripty (as you pointed out).

scripts/test/jest.sh
CMTegner: I just went for consistency for the sake of getting this task done in a reasonable time (by using what I know). We can definitely revisit this as I'm in no way married to scripty. I've used it on a couple of large projects, and it makes structuring your tasks a bit easier, but if we don't see the need then I say we find a better (simpler) solution.

CMTegner:
I amended this PR with some changes to the UMD build. We now no longer do any custom work in this repo, instead we defer the work to [youemdee](https://www.npmjs.com/package/youemdee).

PR #201, opened 2017-01-15
Do not highlight when the event's keyCode is NOT `13`

izumin5210:
For example, when we hit the Enter key to convert Japanese Kana to kanji, KeyboardEvent's keyCode is `229`.

the following screencapture is current behavior when we hit the enter key to convert Japanese Kana to kanji:

[https://gyazo.com/48bdfec2b8014cacb1c7e5ca6881a3d4](https://gyazo.com/48bdfec2b8014cacb1c7e5ca6881a3d4)

Currently, we lose existed text after converting Japanese.

And next screencapture is corrected behavior:

[https://gyazo.com/85fa19b54956400c34fb81a0aaa0615b](https://gyazo.com/85fa19b54956400c34fb81a0aaa0615b)

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -170,11 +170,13 @@ let Autocomplete = React.createClass({
       }
       else if (this.state.highlightedIndex == null) {
         // input has focus but no menu item is selected + enter is hit -> close the menu, highlight whatever's in input
-        this.setState({
-          isOpen: false
-        }, () => {
-          this.refs.input.select()
-        })
+        if (event.keyCode === 13) {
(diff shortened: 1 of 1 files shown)


CMTegner:
Hi! Sorry for leaving this issue open for so long. It looks like you're not using `props.autoHighlight` in your example. I'm just wondering if this should also be applied to the branch below (https://github.com/reactjs/react-autocomplete/blob/master/lib/Autocomplete.js#L275-L291).

CMTegner:
A more general fix has been released in v1.7.1. Please test and verify that it works as intended.

PR #202, opened 2017-01-22
Fix wrapperProps typo in README and comment

golmansax:
I believe `wrapperStyle` is a shorthand for `wrapperProps` and not `inputProps`, but correct me if I'm mistaken.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -84,7 +84,7 @@ let Autocomplete = React.createClass({
      */
     wrapperProps: React.PropTypes.object,
     /**
-     * This is a shorthand for `inputProps={{ style: <your styles> }}`.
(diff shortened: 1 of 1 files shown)


CMTegner:
Good catch!

We generate `README.md` from `README-template.md` via `npm run generate-readme`. Would you mind reverting the two files and applying the change to `README-template.md` instead? Thanks!

golmansax:
@CMTegner I looked in `README-template.md`, but it looks like it gets the API documentation from the source files.  So I modified the change to only update the source file, and tested it by running `npm run generate-readme`.

Let me know if that works!

CMTegner:
Sorry, I got myself confused there for a sec. You're completely right, thanks for the fix!

Issue #192, opened 2016-12-09
Bug: Menu gets triggered by ARROW LEFT and ARROW RIGHT keys


CroniD:
Hi there,

currently if someone press ARROW LEFT or ARROW RIGHT key while the focus is on the input field, the highlight in the menu disappear. Tested with the static data example. It doesn't matter if there is some text in the field or not, I didn't inspect this issue any further, there might be more related background issues.

CMTegner:
Thanks for letting us know! This was fixed in 5d61876fa5a96114d65dbfa3a0a85e3257a415ca which is a part of 1.4.4.

PR #205, opened 2017-02-01
[fixed] Keep highlighted item when pressing arrow left and right keys.

ming-soon:
Refs #192


CMTegner:
Hi! Thanks for putting this together. After looking over the issue I discovered that the problem isn't contained to only left/right arrow keys, but to any key press that doesn't modify `input.value`. I created a general fix in 5d61876fa5a96114d65dbfa3a0a85e3257a415ca which is a part of 1.4.4.

PR #207, opened 2017-02-06
Always render menu dom element to enable transition animation

bfncs:
Currently the menu is not rendered at all when it is not opened. This means each time the open state changes, there is a DOM element addition/removal. Animation this state transition with CSS transitions is therefore not possible.

It would be great to keep this decision to the user. In the pull request, instead of not calling renderMenu at all, it is always called (and thus rendered), but adds `display: hidden;` to the node in the default implementation.

This would be a breaking change, I'm happy to refactor if you have an idea to keep the impact on legacy implementations low.


CMTegner:
Hi! This is a great idea. We probably won't go ahead with it right away (since it's a breaking change), but it'll probably be included in v2. I've already been [tinkered with it](https://github.com/reactjs/react-autocomplete/blob/experimental-rewrite/lib/Autocomplete.js#L73).

bfncs:
Thanks for considering the PR. I understand that this has to wait for the next major version to keep backwards compatibility. Would be great to see it there.

PR #211, opened 2017-02-20
Set ignore blur based on mouse position.

fitz0019:
On Windows where the scrollbar is displayed permanently when the contents is scroll-able, clicking the scroll bar (to start dragging) triggered the blur of the input. This caused the menu to close.

The issue doesn't occur on Mac and you are able to click/drag the scrollbar. I guess Mac doesn't trigger the blur event when clicking part of the system UI.

My solution is to enable/disable the ignoring of the blur based on if the mouse has entered/left the menu. Clicking the menu item would still run and hide the menu as normal. After the changes Mac still behaves as normal.

The other few lines of my commit are removing trailing commas so the code matches the rest.

Hope this makes sense and is okay.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -367,16 +367,20 @@ let Autocomplete = React.createClass({
         onMouseDown: () => this.setIgnoreBlur(true), // Ignore blur to prevent menu from de-rendering before we can process click
         onMouseEnter: () => this.highlightItemFromMouse(index),
         onClick: () => this.selectItemFromMouse(item),
-        ref: `item-${index}`,
+        ref: `item-${index}`
       })
     })
     var style = {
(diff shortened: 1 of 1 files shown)


CMTegner:
Which browser are you experiencing this in? Seems to work as expected in Edge on Windows 10. I'm not saying it's not a problem, just want to be able to verify before we deploy a fix.

fitz0019:
@CMTegner I was testing on IE 11. My mistake for not mentioning that - I had assumed it was the way Windows handled click events on the scroll bar compared to Macs.

It appears @pikelnys has done a better job of explaining the issue within his PR. The only difference between our PR's is the trailing commas I have removed.

CMTegner:
Superseded by #222.

PR #222, opened 2017-04-06
Ignore blur while hovering over menu

pikelnys:
The current `_ignoreBlur` logic does not account for clickable items rendered inside a given `renderMenu` function; it only accounts for items rendered by `renderItem`. This change would ensure that the blur is ignored when clicking anywhere inside the menu.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -379,7 +379,11 @@ let Autocomplete = React.createClass({
       minWidth: this.state.menuWidth,
     }
     var menu = this.props.renderMenu(items, this.props.value, style)
-    return React.cloneElement(menu, { ref: 'menu' })
+    return React.cloneElement(menu, { 
+      ref: 'menu',
+      onMouseEnter: () => this.setIgnoreBlur(true),
+      onMouseLeave: () => this.setIgnoreBlur(false)
+    })
   },
 
   handleInputBlur () {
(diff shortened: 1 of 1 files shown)


pikelnys:
Just saw this PR by @fitz0019, which does the exact same thing https://github.com/reactjs/react-autocomplete/pull/211

To elaborate on the issue, the blur event occurs when you click anywhere inside the menu that was not rendered with the `renderItem` method.

pikelnys:
I just found an edge case with the first commit: when clicking on the menu the input loses focus, so subsequent clicks outside the menu do not trigger a blur. 

I tried to fix it with the subsequent commits, but realized that it led to some unexpected behavior.

jcarter:
Ran into this issue today too

CMTegner:
@pikelnys Thanks for looking into this! I've pushed [a fix](53f229abb7de31fbbcece4285092eec5858a6626) for the problem, and I think I managed to handle the [edge cases](086b74f6cf6c6b5a1c739c7c990772589144a4ba) you mentioned.

Please give 1.5.3 a try and see if it solves the problem.

jcarter:
1.5.3 worked for me. Thanks!

CMTegner:
Just a heads-up to anyone still subscribing to this issue: I just reworked the focus management (released in 1.5.9), so if you start seeing a regression to this fix, or anything else unexpected, please let me know!

PR #225, opened 2017-04-17
Removes the usage of createClass and retrieving prop types from the main React package

luislikescoffee:


nightwolfz:
@luislikescoffee There are conflicts

Issue #223, opened 2017-04-09
createClass, findDOMNode, and string refs, oh my!


fredguest:
Looking at the source, it's a little surprising to see the use of these methods:

- **createClass**
https://github.com/facebook/react/issues/8854
"To reduce the bundle size, we need to remove the APIs that we don't recommend, such as createClass and React.DOM.* helpers."
- **findDOMNode**
https://facebook.github.io/react/docs/react-dom.html#finddomnode
"findDOMNode is an escape hatch used to access the underlying DOM node. In most cases, use of this escape hatch is discouraged because it pierces the component abstraction."
- **string refs**
https://facebook.github.io/react/docs/refs-and-the-dom.html#legacy-api-string-refs
"If you worked with React before, you might be familiar with an older API where the ref attribute is a string, like "textInput", and the DOM node is accessed as this.refs.textInput. We advise against it because string refs have some issues, are considered legacy, and are likely to be removed in one of the future releases."

I see there is an "experimental-rewrite" branch https://github.com/reactjs/react-autocomplete/tree/experimental-rewrite but it doesn't seem to address any of these issues.

What is the reasoning behind the continued use of these discouraged and soon to be deprecated or extracted APIs?

sprjr:
The reasoning is most likely that the project has been 'stable' for awhile and mostly maintained by anyone who needs to drive things forward. I've not been able to carve out much time to invest here, although we can add anyone willing to the project (I believe I can or @ryanflorence can) both here and on npm.

The APIs you listed above have just been part of the module pre-dating their deprecation is all :)

fredguest:
Ok cool that makes sense. I wasn't sure if there was something philosophical behind it because this library seems pretty actively maintained and that `experimental-rewrite` branch is pretty new.

luislikescoffee:
@fredguest I've opened a merge request that removes createClass and getting propTypes from the main react package (https://github.com/reactjs/react-autocomplete/pull/225)

donysukardi:
@luislikescoffee Sorry, I didn't know someone else was working on it too. In addition to React.createClass and React.PropTypes, I also refactored string refs to use refs assignment instead.

luislikescoffee:
@donysukardi No worries, just happy to see it's getting fixed!

donysukardi:
I'm still working on updating the test suite due to the refs changes though.

silicakes:
@luislikescoffee @donysukardi - both of your PRs have conflicts due to the recent update.

[...]

PR #226, opened 2017-04-18
[fixed] Refactor React.createClass, PropTypes, string refs, and fix fakeRequest

donysukardi:
This work is based on #223 and also fixes issue with fakeRequest when `value === ''` and callback doesn't get called.


donysukardi:
Fix has been implemented, which makes PR no longer valid.

PR #231, opened 2017-05-06
[fixed] Upgraded React.createClass(...) to ES6 classes

ryannjohnson:
This PR reconciles https://github.com/reactjs/react-autocomplete/pull/225 and the current master branch.

React complains in the console because it's depreciating its `React.createClass(...)` syntax and separating its `PropTypes` module to its own package. This PR updates that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ We try to follow [http://keepachangelog.com/](http://keepachangelog.com/) recomm
 ## [Unreleased]
 _(add items here for easier creation of next log entry)_
 
+## [1.5.2] - 2017-0511
+### Fixed
+- Prevent onFocus and onBlur when selecting an item (#229)
+
 ## [1.5.1] - 2017-04-27
 ### Fixed
 - Remove logic that selected highlighted item on input click. This was no longer desired after typeahead was removed.
@@ -163,7 +167,8 @@ v0.1.0 - Wed, 12 Aug 2015 19:22:26 GMT
 
 -
 
-[Unreleased]: https://github.com/reactjs/react-autocomplete/compare/v1.5.1...HEAD
+[Unreleased]: https://github.com/reactjs/react-autocomplete/compare/v1.5.2...HEAD
+[1.5.2]: https://github.com/reactjs/react-autocomplete/compare/v1.5.1...v1.5.2
 [1.5.1]: https://github.com/reactjs/react-autocomplete/compare/v1.5.0...v1.5.1
 [1.5.0]: https://github.com/reactjs/react-autocomplete/compare/v1.4.4...v1.5.0
 [1.4.4]: https://github.com/reactjs/react-autocomplete/compare/v1.4.3...v1.4.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ We try to follow [http://keepachangelog.com/](http://keepachangelog.com/) recomm
 ## [Unreleased]
 _(add items here for easier creation of next log entry)_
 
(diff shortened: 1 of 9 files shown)


guillaumepotier:
I'd love to see this PR merged. But it seems 1) it is not synced with master anymore and 2) it brings more features than React update.

Could you please fix that to hope having it merged soon ?

Thx

ryannjohnson:
@guillaumepotier It shouldn't have more features than the ES6 classes update. The git diff does look wonky though, since I pulled `keyDownHandlers` into the constructor to keep it as an object property. Maybe there's a better way?

guillaumepotier:
No you're right. The wonky git diff made me think at first glance you added `keyDownHandlers` new features but you're right. Should be good, pulled your branch in my project and it works like a charm, without nasty React console error warning 👍

CMTegner:
Hi! I really appreciate the effort here! I finally had time to sit down and do the required testing and research on this subject. I've published 1.5.5 which no longer uses `React.createClass` or `React.PropTypes`. I feel like I owe you an explanation since it seems like I just ignored your work and did the same thing as you. Here's my after-the-fact feedback:

* PRs are best when they're small and do one specific thing. This PR contains two important changes that deserve to be reviewed in their own right, as well as a smattering of other things (like changing CJS `require` statements to ES `import`).
* Similarly, keeping commits small and focused really helps speed up the review process. Ideally a commit should only do a single thing. I usually start by reviewing each commit individually, and do a final pass through using the full diff view ("Files changed" tab).
* Another trick you can do to help the reviewer is to keep whitespace changes in [separate commits](00eb7f8961749eb891b0c7240c9955576f14654e). I do a lot of reviewing on the command line, so when I see a commit named "whitespace" or "indentation" and `git diff --ignore-all-space` returns nothing, I know I can safely move on to the next one.
* The changes made to `keyDownHandlers` made the diff much harder to read than it had to, as well as introducing a (small) performance regression. Instead of creating each of the functions in `keyDownHandlers` every time `Autocomplete` is instantiated, I opted for [a solution](https://github.com/reactjs/react-autocomplete/blob/master/lib/Autocomplete.js#L254) where they are available as a static class property, then invoked with the correct lexical scope by using `Function#call` (this was already in place).
* Another small regression can be found in `render`, where the event handlers are being bound for every render. This can be avoided by binding the event handlers [in the constructor](https://github.com/reactjs/react-autocomplete/blob/master/lib/Autocomplete.js#L157-L163).
* The prop-types dependency [should be listed in `dependencies`](https://www.npmjs.com/package/prop-types#how-to-depend-on-this-package). One would think that the natural place would be `peerDependencies`, but I don't blame the React team for wanting to avoid the inevitable complaints that follow.
* The change to prop-types needed a couple of alterations that were not at all obvious. The first was to [exclude the dependency](https://github.com/reactjs/react-autocomplete/commit/c406c4fa84417e3e16b254a4a68bcb2d377b4d7b#diff-566b7a7ceb057cd1ec64691464bdd26c) in the UMD build. The second was to update react-dogcen so the API documentation generator would continue to work.

I hope this can be taken as constructive feedback and not elitism or anything like that. Again, I really appreciate the work you've put in. I figured the path of least resistance was to just push my own changes, since it seemed like it was more important to get the changes out than to spend more time with PRs.

ryannjohnson:
@CMTegner Lol, this isn't elitism at all and is a great resource to learn from. I'd actually like to thank _you_ for taking the time to point out how this PR (and programming) could have been done better. I'll be referring to this next time I decide to contribute to an open repo.

CMTegner:
Great, glad I could be of help! :)

PR #234, opened 2017-05-10
Add placeholder prop and placeholder.

brettdewoody:
Adds a placeholder prop for adding a placeholder to the input.


CMTegner:
Already available via [`inputProps`](https://github.com/reactjs/react-autocomplete/#inputprops-object-optional).

Issue #237, opened 2017-05-22
Arrow Keys with disabled entries


rssteffey:
Arrow keys up/down continue to select disabled menu items (in my particular case, category headers) even though the highlight doesn't render.  Pressing enter while over these will still allow selection of the disabled items

CMTegner:
Hi Shawn! Could you please provide a reduced test case so we can debug this problem?

There's an example of how to add disabled menu items [here](https://reactcommunity.org/react-autocomplete/custom-menu/). Maybe you can compare it to your code and see if there are any obvious differences that may cause problems.

rssteffey:
Sorry, Christian.  I've actually inherited this project and had assumed that the "disabled" property on our menu items was the proper method.

By my understanding of that example the only "correct" way to create category headers is by rendering an additional div based on a condition hit while parsing through _renderItems()?

Our current implementation sets "disabled" styling based on a condition within the individual _renderItem() method instead, since our category headers are actually objects within our results list.  Should that method only be called on actual selectable items only?

rssteffey:
@CMTegner 

I've reached a solution that I would love to incorporate into the repo, but if the general consensus is that the functionality doesn't fit then I can just continue with the forked version.

The root of our issue is that we have category headers returned with the rest of the results in our autocomplete list.  These headers need to display regardless of whether any of their category's items were returned, so we can't [cleanly] generate them within our parsing of list content, they need to be part of the items array themselves.

![screen shot 2017-05-25 at 9 53 23 am](https://cloud.githubusercontent.com/assets/4682038/26455496/091da174-4130-11e7-93d8-172553472392.png)

I've added an optional function isItemDisabled() to the Autocomplete component, very much like the existing shouldItemRender() check.  This function currently only gets called within the arrow key events, and is just used to identify and skip non-selectable items.

![arrow_keys_example](https://cloud.githubusercontent.com/assets/4682038/26456687/74a762b0-4133-11e7-843a-7f95ee9db1ed.gif)
If this seems like a good fit for the project, let me know and I'll submit the request.  I'm sure others have run into similar needs at some point. 

_(I am a bit concerned about the function name, so if there are any clearer suggestions let me know.  I think_ "isItemDisabled" _might imply automatic styling and not allowing clicks, which isn't true on my implementation here.  I've only affected the arrow key controls)_

CMTegner:
[...]

PR #241, opened 2017-05-26
[added] Add isItemSelectable prop function  (#237)

rssteffey:
This adds a bool function prop of isItemSelectable, allowing for category headers or other extraneous information to be presented within the list items. (#237)

This function evaluates on any item that is attempted to be selected (via mouse or keyboard) and prevents these event handlers from occurring (or skips over the item when using arrow key navigation) if the function evaluates false.

By default this function returns true for all items.

_(Request is divided into two commits, the first only adds the functionality and the second is the new example page for Categorized Menu)_

### DIFF
--- a/examples/categorized-menu/app.js
+++ b/examples/categorized-menu/app.js
@@ -1,17 +1,20 @@
 import React from 'react'
 import DOM from 'react-dom'
 import Autocomplete from '../../lib/index'
-import { styles, fakeCategorizedRequest } from '../../lib/utils'
+import { fakeCategorizedRequest } from '../../lib/utils'
 
-let App = React.createClass({
+class App extends React.Component {
 
-  getInitialState() {
-    return {
+  constructor(props) {
+    super(props)
+    this.state = {
       value: '',
-      unitedStates: fakeCategorizedRequest(''),
+      unitedStates: [],//fakeCategorizedRequest(''),
       loading: false
     }
-  },
+    this.requestTimer = null
+    this.renderItems = this.renderItems.bind(this)
+  }
 
   render() {
     return (
@@ -24,7 +27,7 @@ let App = React.createClass({
         <label htmlFor="states-autocomplete">Choose a state from the US</label>
         <Autocomplete
           value={this.state.value}
-          inputProps={{ name: 'US state', id: 'states-autocomplete' }}
+          inputProps={{ id: 'states-autocomplete' }}
           items={this.state.unitedStates}
           getItemValue={(item) => item.name || item.header}
           onSelect={(value, state) => this.setState({ value, unitedStates: [state] }) }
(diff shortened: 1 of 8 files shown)


rssteffey:
(side note, did not update the README because I saw a script that appeared to autogenerate it based off my function header comments.  If that was wrong, just let me know)

lib/Autocomplete.js
```diff
@@ -254,13 +264,17 @@ class Autocomplete extends React.Component {
...
+
       const { highlightedIndex } = this.state
-      const index = (
-        highlightedIndex === null ||
-        highlightedIndex === itemsLength - 1
-      ) ?  0 : highlightedIndex + 1
+      let index = highlightedIndex
+
+      for(let i = 0; i < items.length ; i++) {
+        index = (index === null || index === items.length - 1) ?  0 : index + 1
```
CMTegner: The `index === null` check only needs to be done once, so it can be moved to where `index` is declared and initialized, which will allow you to remove the branching entirely: `index = (index + 1) % items.length`.

lib/Autocomplete.js
```diff
@@ -270,13 +284,17 @@ class Autocomplete extends React.Component {
...
+
       const { highlightedIndex } = this.state
-      const index = (
-        highlightedIndex === 0 ||
-        highlightedIndex === null
-      ) ? itemsLength - 1 : highlightedIndex - 1
+      let index = highlightedIndex
+
+      for(let i = 0; i < items.length ; i++) {
+        index = (index === null || index === 0) ?  items.length - 1 : index - 1
```
CMTegner: If you want to use modulo here, remember to add items.length so you don't go into the negative number range when looping around: `index = (index - 1 + items.length) % items.length`.

lib/Autocomplete.js
```diff
@@ -254,13 +264,17 @@ class Autocomplete extends React.Component {
...
       const { highlightedIndex } = this.state
-      const index = (
-        highlightedIndex === null ||
-        highlightedIndex === itemsLength - 1
-      ) ?  0 : highlightedIndex + 1
+      let index = highlightedIndex
+
+      for(let i = 0; i < items.length ; i++) {
+        index = (index === null || index === items.length - 1) ?  0 : index + 1
+        if (this.props.isItemSelectable(items[index])) break
```
CMTegner: What happens if there are no selectable items?

examples/categorized-menu/app.js
```diff
@@ -0,0 +1,79 @@
+import React from 'react'
```
rmcauley: Hi @CMTegner, I'm trying to think of a way to rewrite the Custom Menu example for this to help this PR along (it's functionality I need for my project as well) and struggling.

The Custom Menu example creates extra elements in the DOM tree based on a condition (the headers do not exist in the list of `items`), where the Categorized Menu example has the categories as part of the returned list from the server and thus as part of the `items` list.

That core difference sets them apart in my mind, even if the end result (from a UX standpoint) remains the same: unselectable headers.

rssteffey: This example seems to be the last outstanding feature since rmcauley's changes.  Let me know if you have any more suggestions, @CMTegner 

I guess the example really comes down to a question of why "Custom Menu" was created in the first place.  Either to demo unselectable headers (we can replace it entirely) or to demonstrate an example of custom rendering (probably worth keeping)

rssteffey:
Hey @CMTegner ! Sorry for the delay.  I've actually gone through and updated the logic to be in line with the changes you made to highlighting functionality back in June.
That should cover all the conflicts that have arisen since then.  I still believe the Custom Menu showcases a fundamentally different use case than the new Categorized one does (custom rendering vs. the boolean isSelectable) so I'm of the opinion we should retain both.

Let me know if you have any more concerns

midgethoen:
I'm in need of displaying disabled items too.
Is there anything I can do to help merge this functionality? or is this available in v2 soon?

ty :)

jcarter:
I second @midgethoen on needing to display disabled items too. This would be fantastic!

rssteffey:
Any chance of a status check on this?  We've managed to work around it for the last several months but that window is getting more difficult to maintain.

yussinsharp:
This would also help my use-case for displaying "No Suggestions Found" components and some "Fetching..." components as the menu will not show if there are no suggestions.

CMTegner:
Sorry for taking so long, I haven't had much time to tend to this project lately. I rebased your changes, fixed a few bugs, added a few more tests, and rewrote the "custom menu" example to use `isItemSelectable`. Released in 1.8.0. Hope you don't mind me hijacking your branch. For my own sanity I try to keep the commit history as clean as possible, and this PR was getting a bit cluttered with all the merges and fixes.

Thanks for the work!

Issue #153, opened 2016-08-01
Menu does not closing after selecting one. onliy ie


kortv:
When you try to select one item from the menu, the menu is not locked. For me this is a problem on the IE11 in other browsers work fine.
This can be demonstrated in ie11 using the examples linked on the site: 
https://reactcommunity.org/react-autocomplete/async-data/

Zara603:
I have the same problem, only in IE.

kortv:
[Here](https://github.com/reactjs/react-autocomplete/pull/154/files) simple solution of this problem.
So i have change Autocomplete component for my needs, and it's work fine.
That's happening cause IE focus() work, a little bit different, not like in over browsers

CMTegner:
Thanks for brining this to our attention! I solved the problem in a slightly different way to prevent having to introduce a timeout. I tested it in IE11 on Windows 8.1 and it seems to work fine. A new release should be on npm in a couple of days. /cc @sprjr

Zara603:
@CMTegner Thanks for the fixation. lodash has been removed from dependencies, though. I was getting ReferenceError: _ is not defined.

Zara603:
@CMTegner I still have this problem on IE that the item is not selected.

CMTegner:
>lodash has been removed from dependencies, though. I was getting ReferenceError: _ is not defined.

The lodash dependency was removed over two months ago. Which version of react-autocomplete are you running?

>I still have this problem on IE that the item is not selected.

Do you mean that the menu isn't closing? The fix hasn't been released on npm yet, so I assume you're building from master, correct?

Zara603:
@CMTegner after I've updated from 0.2.1 to 1.1.0 I'm getting ReferenceError: _ is not defined

CMTegner:
You're going to have to open a new issue and provide a reduced test case for us to help you further.

dennisat:
I still have this issue. in IE11 and Edge only. version 1.7.2

PR #246, opened 2017-06-08
[fixed] Fix dropdown not closing on selection in IE

rssteffey:
Issue #153

I suspect a change since this ticket was first opened has broken it again.  Example pages do not close menu on selection in IE11.  This adds the fix back into the inputBlur handler

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -430,6 +430,7 @@ class Autocomplete extends React.Component {
 
   handleInputBlur(event) {
     if (this._ignoreBlur) {
+      this.setIgnoreBlur(false)
       return
     }
     this.setState({


CMTegner:
Hi Shawn! Thanks for bringing this to our attention, and sorry for the late response. I've been beating my head against the wall, trying to solve the "focus management" problem for the past 4-5 days. I've finally managed to release something (1.5.9) that seems to handle all of the edge cases.

This specific issue should of course never have regressed, but since I didn't add a test for it the first time I fixed it, I guess I had it coming ¯\_(ツ)_/¯

The reason why I didn't accept your fix is that it caused the logic which is supposed to preserve focus and keep the menu open, to become unsynced and stop working properly. You could provoke this by clicking on a non-item element within the menu (e.g. a scrollbar or some padding on the menu), and then try to select an item. The second blur would case the menu to close before the item got selected, since the `ignoreBlur` flag had been reset.

You can read more about how I solved this and related problems in the [commit message](https://github.com/reactjs/react-autocomplete/commit/ec0c6be23285d286fcb9a4ba4da5f5c9960d0179).

dennisat:
I still have this issue! how is that possible?
Any help is appreciated! Thanks.
Although the https://reactcommunity.org/react-autocomplete/async-data/ is working, using the latest version (1.7.2) I have this problem only in IE and Edge. What version does this link use?

PR #252, opened 2017-06-21
[added] Auto select highlighted value on focus lost

### LINKED ISSUE #251: Autoselect (focus lost)
For our usecase we actually need that blur or focuslost causes the value in dropdown to be preselected when user leaves the field.

The idea is that some users when item is highlighted may think it is preselected.

SergeyShvager:
Added new `props.autoSelect`. If `autoSelect=true` highlighted value will be selected when user clicks outside of Autocomplete component.

Closes #251 issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,12 @@ Default value: `true`
 Whether or not to automatically highlight the top match in the dropdown
 menu.
 
+#### `autoSelect: Boolean` (optional)
+Default value: `false`
+
+Whether or not to automatically select highlighted value when user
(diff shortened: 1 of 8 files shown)


matiwinnetou:
👍

CMTegner:
This was released in v1.7. I renamed the prop to [`selectOnBlur`](https://github.com/reactjs/react-autocomplete#selectonblur-boolean-optional) and removed the example. I'll add it back once I'm done rewriting the examples.

PR #258, opened 2017-07-18
Fix binding error with preact-compat

bgelineau:
When using ``preact-compat``, the updater function passed to the ``setState`` method is not called in the context of the class. Thus, ``ensureHighlightedIndex`` and ``maybeAutoCompleteText`` crash when calling ``this``, which is undefined.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -190,10 +190,10 @@ class Autocomplete extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     if (this.state.highlightedIndex !== null) {
-      this.setState(this.ensureHighlightedIndex)
+      this.setState(::this.ensureHighlightedIndex)
     }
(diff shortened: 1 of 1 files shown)


CMTegner:
Hi! Any chance you can bind these methods in the constructor using `Function#bind` directly instead? We'll probably move away from transpiling stage 0 proposals in the future (stage 2/3 should be enough), and this particular proposal hasn't seen any movement in a while.

bgelineau:
Hi! No problem, here is the new version without the bind operator.

Indeed, from what I've seen (and jest), your code already works with stage 3 + class properties.

CMTegner:
Thanks! Released in v1.5.10. Let us know if you find any other incompatibilities so we can try to keep `Autocomplete` working for both React and React-like libraries.

bgelineau:
Thanks.

Issue #247, opened 2017-06-13
add custom input component


jallwine:
It would be great to have more control over the input component. What if you allowed an input prop that could be a custom component that wraps an &lt;input&gt; tag? If you don't provide one, it defaults to <input> like it currently does. 

I've been playing around with the idea in my own fork: https://github.com/jallwine/react-autocomplete/blob/master/lib/Autocomplete.js

It's on npm as react-autocomplete-custom-input, currently.

As long as the custom input component provides a ref to the actual input node (ref={ input => this.input = input} on the input component) and accepts all the props that Autocomplete uses on the input, it seems to work. This was a quick test, so any thoughts or concerns about doing it this way would be great.

CMTegner:
Do you have a use-case for this?

>As long as the custom input component provides a ref to the actual input node (ref={ input => this.input = input} on the input component)

`ref` doesn't get passed to the child component, it's only visible to the owner. For this to work the custom input component would have to expose all the native `<input>` APIs (`focus`, `select`, etc.), like `Autocomplete` does by proxying to `refs.input`. It would also need to be a DOM element, or else we'll have to use `findDOMNode` like we do with the items and menu. In other words: it would have to do a lot of what `Autocomplete` already does.

The other argument is that `Autocomplete` is quite tightly coupled to the behaviour of the native `<input>`, especially when it comes to focus management. I fear that any deviation in your custom input component could cause unexpected behaviour.

We want to make `Autocomplete` as close to the native `<input>` as we can get. With this in mind, it may be better to pass `Autocomplete` in place of an `<input>`, instead of passing a custom input component to `Autocomplete`.

I think the main argument for this type of behaviour would be to detach `Autocomplete` from ReactDOM, so it could be used in other environments (e.g. react-native). I haven't worked enough with inputs in react-native to know if this is even possible, but i think the idea is interesting.

jallwine:
I have a custom Input component that is wrapped in several divs with places for labels and icons and other features. I'd like to hook up autocomplete features to it, but the Autocomplete component owns the input component, so I either need to implement my own autocomplete features into the custom Input component, or integrate the custom pieces into the Autocomplete component.

jallwine:
[...]

PR #262, opened 2017-07-28
feat: Inject Custom Input component type

abdennour:
- New props named "as" to inject custom component type , default is "input"

```js
import MyInput from './MyInput'
<AutoComplete as={MyInput} {...otherProps} />
```

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -28,6 +28,7 @@ function getScrollOffset() {
 class Autocomplete extends React.Component {
 
   static propTypes = {
+    as: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
     /**
      * The items to display in the dropdown menu
      */
@@ -134,6 +135,7 @@ class Autocomplete extends React.Component {
   }
 
   static defaultProps = {
+    as: 'input',
     value: '',
     wrapperProps: {},
     wrapperStyle: {
@@ -505,10 +507,11 @@ class Autocomplete extends React.Component {
     }
 
     const { inputProps } = this.props
+    const Input = this.props.as
     const open = this.isOpen()
     return (
       <div style={{ ...this.props.wrapperStyle }} {...this.props.wrapperProps}>
-        <input
+        <Input
           {...inputProps}
           role="combobox"
           aria-autocomplete="list"
@@ -534,4 +537,3 @@ class Autocomplete extends React.Component {
 }
 
 module.exports = Autocomplete
-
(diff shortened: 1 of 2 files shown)


abdennour:
It resolves #247  :)

CMTegner:
@abdennour seems like you've ignored most/all of the discussion in #247. Simple testing proves that this doesn't work due to missing `ref` support (your test is too naive: it only tests that the provided component is rendered, not that it behaves as expected).

This feature will not likely be implemented before some solid use-cases are presented, and a thorough discussion regarding the API surface has been performed. That was the intention behind keeping #247 open. If you have a strong argument as to why this feature is needed, please present it for discussion there.

Dattaya:
@CMTegner one use case is to be able to use `material-ui/TextField` or `material-ui/Input` underneath (I'm talking about v1.0-beta). `react-autosuggests` allows that as can be seen in [an example in the `material-ui` docs](https://material-ui-1dab0.firebaseapp.com/component-demos/autocomplete). Though in addition to `as` it would require a `refFuncName` property. I did that created a component:

```js
const AutocompleteTextField = (props) => (
  <Autocomplete
    inputComponent={TextField}
    refFuncName="inputRef"
    {...props}
  />
);
```
and here's the result:
ezgif-1-129e49111a

CMTegner:
v1.6.0 includes support for [`props.renderInput`](https://github.com/reactjs/react-autocomplete#renderinput-function-optional) which can be used to the same effect as this PR.

CMTegner:
@Dattaya thanks for the info! You should be able to use `renderInput` to make `Autocomplete` wrap a Material-UI TextField.

Dattaya:
I tried this new API and it works flawlessly with `material-ui` and of course the same holds true for other UI libraries. Thank you, @CMTegner!

CMTegner:
No problem, glad i helped! :)

azcn2503:
Hi all,

When trying to use `renderInput` to render a Material UI TextField, I get errors regarding accessing the input ref. It seems that this is because AutoComplete is trying to access `getBoundingClientRect()` of the TextField component rather than TextField's `input` ref, which is accessible at `this.refs.input._el.input`. Any ideas for a workaround? Or should I raise an issue?

CMTegner:
You need to pass `props.ref` to [`inputRef`](https://material-ui-next.com/api/input/):

```jsx
renderInput({ ref, ...props }) {
  return <Input inputRef={ref} {...props} />
}
```

PR #276, opened 2017-09-05
Resolve preact-compat compatibility issue

eve-bright:
I am using react-autocomplete in a preact application using preact-compat (https://github.com/developit/preact-compat) and had an issue with highlighted index on value change. 

This was due to preact-compat using a different queueing mechanism for setState within componentWillReceiveProps which resulted in the old properties (in particular the value) being used when evaluating what the new highlighted index should be. 

I have resolved this by passing the new props explicitly to the ensureHighlightedIndex and maybeAutoCompleteText functions.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -214,10 +214,10 @@ class Autocomplete extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     if (this.state.highlightedIndex !== null) {
-      this.setState(this.ensureHighlightedIndex)
+      this.setState(this.ensureHighlightedIndex(nextProps))
     }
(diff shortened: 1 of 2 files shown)


CMTegner:
Hi Eve! Can you point to the documentation that describes the differences? I've had a look, but I couldn't find anything related. I'm hesitant to accept this change as the main reason for using the "updater" style `setState` API is that you receive the latest state and props so you're always computing the new state based on the latest data. Manually passing state/props defeats the purpose of this technique.

eve-bright:
Hi Christian, I can't find any documentation, this was just what I observed. 

The "updater" style setState API does seem particularly elegant and it's frustrating that it doesn't seem to work with preact-compat. I will close the PR for now and see if I can understand better why it does not work.

PR #281, opened 2017-10-04
[fixed] maybeScrollItemIntoView exception when itemNode is not found

faviktor92:


CMTegner:
It's difficult to know why you want to perform this change, since you've provided no context, but I'm assuming it's because you've implemented a `renderMenu` function which doesn't always return every item to be rendered. This edge case is already [documented](https://github.com/reactjs/react-autocomplete#rendermenu-function-optional):

>Ensure the returned tree includes every entry in items or else the highlight order and keyboard navigation logic will break.

PR #285, opened 2017-10-16
[changed] isOpen function treats null or undefined prop as not being set

acunniffe:
Some users may want to use the isOpen override selectively depending on the app state. Right now even if you set it as undefined or null the code treats that as a definitive false. 

Currently:
true is treated as true
false is treated as false
null is treated as false
undefined is treated as false

With this PR:
true is treated as true
false is treated as false
undefined is treated as [not set]
null is treated as [not set]

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -525,7 +525,8 @@ class Autocomplete extends React.Component {
   }
 
   isOpen() {
-    return 'open' in this.props ? this.props.open : this.state.isOpen
+  	if ('open' in this.props && this.props.open !== undefined && this.props.open !== null) return this.props.open
+  		else return this.state.isOpen
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


CMTegner:
Hi Aidan! I agree, this change is probably sensible. The original thought was that `props.open` would be either managed, or unmanaged, like how `props.value` works in React's implementation of `<input>`. That's also why we implemented [`props. onMenuVisibilityChange`](https://github.com/reactjs/react-autocomplete#onmenuvisibilitychange-function-optional), so that you could mirror `Autocomplete`'s open state when you didn't want to override it.

Sadly, changing this logic as proposed would probably break a few implementations. It's very easy to pass `undefined`/`null` instead of `false` by accident, and not everyone is good at ensuring there are no prop type violations. I'll hold off on this until v2.

PR #301, opened 2017-12-20
New default behavior onMouseOut

vladimir-konnov:
Hello all!

I would like to propose new default behavior when user moves mouse out of the dropdown box: remove highlighting from last selected item. People are asking why the highlighting stays when you move mouse out of options list.
I think it was done on purpose for the sake of selectOnBlur functionality. But now there's no ability to change this behaviour, no option in props list. So I'm proposing to remove highlighting by default when selectOnBlur is off.

Thank you

P.S. Though I proposed this as a default behavior, you'd probably like to add an option into props list to handle it explicitly. I'm also ok with that

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -433,6 +433,7 @@ class Autocomplete extends React.Component {
       )
       return React.cloneElement(element, {
         onMouseEnter: () => this.highlightItemFromMouse(index),
+        onMouseOut: () => { if (!this.props.selectOnBlur) this.highlightItemFromMouse(null) },
         onClick: () => this.selectItemFromMouse(item),
         ref: e => this.refs[`item-${index}`] = e,
       })


vladimir-konnov:
Hi CMTegner! I proposed this code in order to fix the case when user moves the cursor out of a menu item, the same like it highlights a menu item `onMouseEnter`.
Well, from performance point of view I agree with you, it would be better to remove highlighting on mouse leaving the menu itself, not just a menu item. For the first glance it should work ok if you move the handler from menu item to menu itself.

viswanathm:
Hi, I agree with Vladmir. I am using this control and facing some problem of not removing highlighted index on mouse out. 

And moving mouseOut event from menu item to menu would improve the performance too.

When will this pull request be merged? I can probably move the mouseOut to menu from menuItem and raise new pull request.

It would be great if this can be addressed at the earliest.

PR #303, opened 2018-01-10
Added setIgnoreBlur(false) to Enter-event

Fralleee:
This should fix issue #304


CMTegner:
Holy edge-case, Batman! Good catch Roland! I took your fix and [adjusted it](https://github.com/reactjs/react-autocomplete/commit/6deffa455ae4ba033f94f0781fa89f03f61acde5) a bit, since the bug is also present in the branch above (the menu may be hoverable even though it has no items).

Thanks!

Issue #309, opened 2018-01-31
autoHighlight not working unless match is beginning of an item


epegzz:
Let's say you have the following items: `Foo`, `Bar`

If you start typing `F` then `Foo` will automatically be highlighted.

But if you instead start typing `o` then `Foo` will not be highlighted.
Therefore, hitting Enter will not select the first item in the second scenario.

CMTegner:
Duplicate of #239.

Issue #239, opened 2017-05-23
autoHighlight is not always highlighting the top match


OliverBroszat:
The top match is only highlighted if it is compared to a String with the exact same beginning.

For Example if I want to search for "123" it will only be highlighted if I enter "1" and it is not being highlighted if I write "2" or "3" in the input field.

CMTegner:
The current implementation is equivalent to `String#startsWith`. We don't have any plans to change this in v1, but it is on our things to consider for future versions.

silicakes:
@CMTegner - i see the line right here:
https://github.com/reactjs/react-autocomplete/blob/master/lib/Autocomplete.js#L352 

```
 const itemValueDoesMatch = (itemValue.toLowerCase().indexOf(
        value.toLowerCase()
      ) === 0)
```

Why is it difficult to change it to something like:

```
 const itemValueDoesMatch = (itemValue.toLowerCase().includes(
        value.toLowerCase()
      ) === 0)
```

CMTegner:
It's not difficult, we just don't intend on changing the behaviour at the moment. When we eventually do, we will want to create something that is flexible enough to cater to different needs, and not just a single use case.

silicakes:
@CMTegner - I've forked the repo, added a predicate support, which is essentially function receiving `itemValue: String` and `inputValue: String` and should return a boolean result,
You can see it here:
https://github.com/reactjs/react-autocomplete/compare/master...madlandev:master 

It's not possible for me to create a fork since I've included the build files for our own modules, but should you desire, I can create one for you guys.

ryansully:
@silicakes Would you be interested in submitting a PR with your changes to implement this? It seems like it should be expected functionality.

BeauCarrel:
I've found a "solution" for my use case, I just want it to select the first in the list by default, so that the user can hit enter and it will select the first in the list, for the getItemValue(item) just return the input query, that way the first always matches the default match and is selected until the user starts navigating with the arrow keys.
It's not pretty but it works for my use case :)

PR #310, opened 2018-01-31
[fixed] #309 - autoHighlight not working unless match is beginning of…

epegzz:
See #309 

The current implementation only auto-highlights an item if the input matches the beginning of the item label.

That's problematic if your `shouldItemRender` implementation allows matching anywhere in the item label, not just from the start.

This PR fixes that.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -375,7 +375,7 @@ class Autocomplete extends React.Component {
       const itemValue = getItemValue(matchedItem)
       const itemValueDoesMatch = (itemValue.toLowerCase().indexOf(
         value.toLowerCase()
-      ) === 0)
+      ) > -1)
       if (itemValueDoesMatch) {
         return { highlightedIndex: index }
       }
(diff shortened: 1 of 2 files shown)


lib/__tests__/Autocomplete-test.js
```diff
@@ -65,7 +65,7 @@ describe('Autocomplete acceptance tests', () => {
   it('should highlight top match when `props.value` changes', () => {
     const tree = mount(AutocompleteComponentJSX({}))
     expect(tree.state('highlightedIndex')).toEqual(null)
-    tree.setProps({ value: 'a' })
+    tree.setProps({ value: 'ri' })
```
epegzz: The item we are matching here is `Arizona`.
In the old test we typed `a`, which put the `Arizona` item first and auto-highlighted it.

But if you typed `ri` then still `Arizona` was put as first item, but it was not auto-highlighted, since `ri` does not match the beginning of `Arizona`.

With the changes in this PR, `Arizona` will now correctly be highlighted.

lib/Autocomplete.js
```diff
@@ -375,7 +375,7 @@ class Autocomplete extends React.Component {
       const itemValue = getItemValue(matchedItem)
       const itemValueDoesMatch = (itemValue.toLowerCase().indexOf(
         value.toLowerCase()
-      ) === 0)
+      ) > -1)
```
epegzz: I'd argue, that `itemValueDoesMatch` is unnecessary altogether.
If an item is displayed because it passed the check of `shouldItemRender` then we can also auto-highlight it if it's the first item in the list.

CMTegner:
The existing logic won't be changed at the moment. See #239 for explanation.

Issue #298, opened 2017-12-04
Don't return a `div` but just an array (React 16)


marcelhageman:
Right now I can't do this in CSS:

`input:focus + label { color: red; }`

Because the output for the react-autocomplete contains the `input` element inside a `div`.

Instead, I'd like to be able to output only an `input` element and render it right beside a `label`, so I can focus the input field and style the next sibling `label` element accordingly.

Since React 16 I believe you can also render as follows:

```javascript
const renderAutocomplete = [
  {this.props.renderInput(...)},
  {open && this.renderMenu()},
  {this.props.debug && ...}
];
```

You'd need to provide a `key` attribute to each child.

The containing `div` could be skipped since it offers no functional meaning, you could make it the default for old-times sake.

```javascript
const { renderWrapper } = this.props;

return (
  renderWrapper
    ? <div style={{ ...this.props.wrapperStyle }} {...this.props.wrapperProps}>
        {renderAutocomplete}
      </div>
    : renderAutocomplete
);
```

Thoughts?

CMTegner:
Sure! At some point we'll drop support for React <16 and this will be how the component functions, but until then this is a good compromise. Create a PR and I'll review it as soon as I have time.

PR #314, opened 2018-02-08
use span instead of div for wrapper-element

flash1293:
This PR uses a span as the wrapping element instead of a div. This makes it possible to use the component inside of elements which prohibit the usage of block-elements as children like paragraphs.

Example:
```
<p><Autocomplete {/*...*/} /></p>
```
Prodcues the error `<div> cannot appear as a descendant of <p>. See ... > Autocomplete > div`.


CMTegner:
Hi Johannes! I think I'd rather wait for something like https://github.com/reactjs/react-autocomplete/issues/298 instead of changing the root node type, as this will likely cause problems for people who rely on the rapper being a `<div>`. If you feel like having an attempt at implementing the that suggestion it would be greatly appreciated.

flash1293:
Fair enough - what about a new prop `renderWrapper` analogous to `renderInput` and `renderMenu`?
It would be a function with the parameters `wrapperProps`, `wrapperStyle`, `renderedInput` and `renderedMenu`, default is something like
```jsx
return <div style={{wrapperStyle}} {...wrapperProps}>
{renderedInput}
{renderedMenu}
</div>;
```
Pros:
* no breaking change
* #298 covered
* even more flexibility for the consumer

Would be happy to adjust the PR if you agree.

CMTegner:
Sounds good! I wonder if maybe we should rather pass in `renderInput` and `renderMenu` in addition to `isOpen`, so you have full control over what you want to render.

flash1293:
@CMTegner I added the `renderWrapper` prop. I'm not sure what you mean by full control. With `open`, `renderMenu`, `renderInput` and now `renderWrapper`, the user should be in complete control of the render-output.

bitionaire:
y no merge?

PR #315, opened 2018-02-08
Fix ArrowUp to properly skip over unselectable items

rssteffey:
Arrow up is broken in 1.8 and won't iterate past unselectable items (category headers in the custom states example).  This fixes it.


CMTegner:
Released in 1.8.1.

PR #316, opened 2018-02-09
Fixing crash when itemNode is not found

rek:
To trigger bug:

Type into the autocomplete field, when results appear, hover the mouse over the results.
Then backspace to remove the text, and the following error appears:

image

This PR adds an easy check to mitigate that.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -255,11 +255,16 @@ class Autocomplete extends React.Component {
     if (this.isOpen() && this.state.highlightedIndex !== null) {
       const itemNode = this.refs[`item-${this.state.highlightedIndex}`]
       const menuNode = this.refs.menu
-      scrollIntoView(
-        findDOMNode(itemNode),
-        findDOMNode(menuNode),
-        { onlyScrollIfNeeded: true }
-      )
+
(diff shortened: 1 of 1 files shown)


CMTegner:
Hi Adam! I can't reproduce this error. Are you sure you're rendering all the items specified in `props.items`?

rek:
Hrmm strange, yeh I think everything is setup right. Did you make sure to keep the mouse cursor over the dropdown results when clearing the input text?

If you still have trouble, I will make a short screen cap.

CMTegner:
Can this be reproduced with any of the examples on http://reactcommunity.org/react-autocomplete/?

rek:
Good point! No I cannot reproduce it there. 

It must be to do with something in my setup. I am thinking now perhaps it is because we make the dropdown box smaller and add a scroll bar, as you alluded to. I will research more tomorrow.

rek:
Ok finally found it, we were missing an 'items' class in renderMenu. 

So perhaps my suggested fix here is not right, but I do feel there is some place we could make a change to the lib so it crashes more gracefully.

Though I don't really have time to look into it, thanks for keeping this open so long, i'll close it now.

PR #337, opened 2018-05-04
Add new optional prop isItemAutoHighlightMatch

### LINKED ISSUE #266: autoHighlight={true} only highlights the first item - if it matches from its beginning
**autoHighlight={true} only highlights the first item if it matches from its beginning**

Here's an example for when it doesn't work:
1. user inputs: "foo"
2. multiple choices for items containing the word "blah blah food" are shown
3. first item isn't highlighted

It does work for the following:
1. user inputs "foo"
2. multiple choices begin with . "food"
3. first item is highlighted

bfncs:
I added a new optional prop isItemAutoHighlightMatch that allows injecting a function to determine whether an item should be autohighlighted. The default prop implementation checks whether the value of the proposed item starts with the search value  ignoring case, so there is no change in the default behaviour.

This would fix #266, #239, #310 and #239 without braking backwards compatibility or making a more involved design in the future less feasible. If this is ok, I'd gladly add some documentation, but I wanted to ask for some comments first. I think this feature is essential for advanced usage of react-autocomplete with own matching implementations.

### DIFF
--- a/lib/Autocomplete.js
+++ b/lib/Autocomplete.js
@@ -25,6 +25,9 @@ function getScrollOffset() {
   }
 }
 
+const startsWithIgnoreCase = (haystack, needle) =>
+    haystack.toLowerCase().indexOf(needle.toLowerCase()) === 0
+
 class Autocomplete extends React.Component {
 
   static propTypes = {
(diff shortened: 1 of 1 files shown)


bfncs:
Please let me know how I can help to get this merged or find a better solution.

zackify:
Really confused why the maintainers don't want to add this. My autocomplete results have other fields that match on them, and the top one should always get highlighted :(

chalda:
having a slightly different issue which also results in items not autohighlighting because of the highlighting function in the lib. this PR would help since this exposes the highlighting logic and i can just reimplement it for my own datamodel

PR #363, opened 2018-10-25
[added] `isClearable` prop

YahyaAhmad:
To render a clear button using fontawesome, it will invoke the `onClear`  prop function when clicked.


CMTegner:
I appreciate the time and effort it took to create this PR, but we do not have any desire to internalise this functionality as you have all the hooks required to do it outside of `Autocomplete`. It's a good idea to propose changes like this via an issue before investing time into implementation.

YahyaAhmad:
Thank you, I understand. This is my first time contributing to a project. I'm still new to this, so forgive my incompetence. Can you please give me some tips on contributing to open-source projects?

CMTegner:
Do not undermine your own skills, you are in no way incompetent! It's not always a straightforward task to identify what you could/should contribute to a project.

I assume the desire to add the "clear data" button came from a real need that you have, in which case you're already off to a good start: you identified a need and tried to satisfy it. Next up you should do some critical thinking and ask yourself questions such as:
* Can the task/action/behaviour be done easily with the APIs/tools already provided?
* Does what you're adding help make the project do what it does better, or is it adding fringe "cruft" that doesn't represent the project's core values?
* Is what you're adding needed by so few users that the cost (increased bundle size, initialization/render time, etc) is not justifiable towards the average users?
* Is the added complexity worth the added maintenance cost?
* Will other users want to modify this feature/behaviour, potentially creating future added complexity to satisfy subsequent needs?

The easiest way to "break through" and get your code landed in open source projects is probably to follow the issues of projects you use/like, and see if there are any low hanging fruits you can contribute to. This could be bugs (look for tags such as "beginner friendly" in larger repositories that do issue triaging) or minor changes/features that are in high demand. Small things such as answering questions (e.g. if you have experience using the project) could go a long way to being included in a project's "inner circle" of trusted contributors, which could help you become included in the project's development process.

Good luck!

YahyaAhmad:
Thank you for answering my question, I will make sure to remember the next time I contribute something.

